PR #1, opened 2026-07-25
feat: Memory-adaptive startup with context fallback

giveen:
## Summary
- Adds an opt-in `--context-fallback` path for `ninfer-serve` models that cannot fit a requested `--max-context` on the target GPU.
- Retries target construction with progressively smaller context lengths when the initial allocation exceeds available VRAM.
- Propagates the realized effective context limit through `EngineOptions`, `LoadSummary`, and serve-side metadata so `/v1/models` reports the true built capacity.

## Test Plan
- [x] Build succeeds after CMake + Ninja rebuild
- [x] `ninfer-serve` accepts new flags: `--context-fallback`, `--context-fallback-min`, `--context-fallback-step`
- [x] Verified on a 27B model with requested 262144-token context, requesting DAO to `227328` tokens instead of hard failing on a partially loaded desktop GPU
- [x] End-to-end request returns real completions on the reduced context (`ttft=22ms`, `prefill=502.7tok/s`, `decode=84.2tok/s`)
- [x] Default behavior unchanged when `--context-fallback` is not provided

## Affected files
- `include/ninfer/types.h`
- `src/runtime/engine/engine.cpp`
- `src/serve/generation_service.cpp`
- `src/serve/serve_options.cpp`
- `src/serve/serve_options.h`
- `src/targets/registry.cpp`

## Notes
- Keeps changes minimal and opt-in so existing benchmarking behavior remains unchanged.
- The fallback already validated against the live desktop GPU with both a smaller context and the oversized 262K target.

### DIFF
--- a/include/ninfer/types.h
+++ b/include/ninfer/types.h
@@ -52,6 +52,9 @@ struct EngineOptions {
     bool enable_vision  = false;
     bool use_cuda_graph = true;
     LoadProgress load_progress;
+    bool allow_context_fallback        = false;
+    std::uint32_t context_fallback_min = 4096;
+    std::uint32_t context_fallback_step = 2048;
 };
 
 struct SamplingParameters {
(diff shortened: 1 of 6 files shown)


Neroued:
@codex review

PR #4, opened 2026-07-28
feat(build): support native Windows builds

pinizhaninov:
## Problem

NInfer's RTX 5090 target builds and runs in Linux/WSL, but the upstream tree depends on POSIX file mapping/direct I/O, process/network APIs, and pkg-config dependency discovery. Those assumptions prevent a native Windows build even though the CUDA execution target remains the same `sm_120a` RTX 5090.

## Design

- Keep the existing Linux/WSL paths unchanged.
- Use vcpkg CMake integration for FFmpeg and curl on Windows.
- Link the static CUDA runtime consistently with the MSVC/vcpkg runtime.
- Implement the artifact reader with `CreateFileMappingW`/`MapViewOfFile` and aligned `FILE_FLAG_NO_BUFFERING | FILE_FLAG_OVERLAPPED` reads.
- Initialize Winsock for media host resolution and use the Windows process API for request IDs.
- Apply the small MSVC portability fixes required by the current tree.
- Document a native PowerShell/Ninja build that selects the newest installed MSVC toolchain.

This changes build/platform support only. It does not change the registered models, artifact format, numerical kernels, `sm_120a` target, CLI/serving protocol, or inference behavior.

## Provenance

The Windows portability implementation adapts work from [`Don-Chad/ninfer-3090` commit `a1c7c2e`](https://github.com/Don-Chad/ninfer-3090/commit/a1c7c2e11c60e8034e11ff1c90944a84db412ace). Don-Chad is credited in the commit trailer and README. This PR retargets that portability layer to the current RTX 5090 upstream; the fork's RTX 3090 kernel retargeting and performance changes are not included.

## Verification

Hardware: NVIDIA GeForce RTX 5090.

Native Windows 11 (`10.0.26100`), CMake 4.0.3, MSVC 19.51.36248, CUDA 13.3.33:

```powershell
cmake -S . -B build-pr-windows -G Ninja `
  -DCMAKE_BUILD_TYPE=Release `
  -DCMAKE_CUDA_COMPILER='C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.3\bin\nvcc.exe' `
  -DCMAKE_TOOLCHAIN_FILE='<Visual Studio>\VC\vcpkg\scripts\buildsystems\vcpkg.cmake' `
  -DVCPKG_TARGET_TRIPLET=x64-windows `
  -DNINFER_BUILD_APPS=ON -DBUILD_TESTING=ON
cmake --build build-pr-windows --parallel
ctest --test-dir build-pr-windows --output-on-failure -j 1 `
  -R 'ninfer_(public_api|artifact_reader|artifact_materialization|row_split_pack_golden|qwen3_6_frontend|request_log|vision_attention)_test'
build-pr-windows\apps\ninfer-serve.exe --help
```

Results:

- clean configure succeeded;
- all 593 build steps succeeded, including both applications and all test binaries;
- all 7 focused tests passed;
- the native `ninfer-serve.exe --help` smoke test passed.

The golden-test dependency was installed offline from uv's existing cache (`safetensors 0.7.0`); `ninfer_row_split_pack_golden_test` then passed. The full serialized GPU suite was not run because an active `ninfer-serve` process on the same RTX 5090 was intentionally left undisturbed; the focused portability tests passed and every test executable built successfully.

WSL2 Ubuntu 24.04, CMake 3.28.3, GCC 13.3.0, CUDA 13.1.115:

```bash
cmake -S . -B build-pr-wsl -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_CUDA_COMPILER=/usr/local/cuda-13.1/bin/nvcc \
  -DNINFER_BUILD_APPS=ON -DBUILD_TESTING=OFF
cmake --build build-pr-wsl --parallel
build-pr-wsl/apps/ninfer-serve --help
```

All 193 application build steps and the WSL `ninfer-serve --help` smoke test passed. No performance claim is made by this PR.

## AI assistance

The adaptation, review, documentation, and validation were performed with OpenAI Codex based on GPT-5. The submitted diff was reviewed against upstream and the source port; the contributor remains responsible for the submitted changes.


Neroued:
Thanks for the PR! I only have a couple of code-organization preferences:

1. Could you please abstract the file access into a platform-independent `ReadOnlyFile` RAII object owned by `core`? For example:

```cpp
class ReadOnlyFile {
public:
    explicit ReadOnlyFile(const std::filesystem::path& path);
    ~ReadOnlyFile();

    ReadOnlyFile(ReadOnlyFile&&) noexcept;
    ReadOnlyFile& operator=(ReadOnlyFile&&) noexcept;

    ReadOnlyFile(const ReadOnlyFile&) = delete;
    ReadOnlyFile& operator=(const ReadOnlyFile&) = delete;

    std::span<const std::byte> mapped_bytes() const noexcept;

    std::size_t read_direct(
        std::uint64_t offset,
        std::span<std::byte> destination) const;

private:
    struct Impl;
    std::unique_ptr<Impl> impl_;
};
```

The implementation could use two platform-selected source files:

```text
src/core/read_only_file.h
src/core/read_only_file_posix.cpp
src/core/read_only_file_win32.cpp
```

CMake should compile only the implementation for the current platform. This keeps Win32/POSIX handles, mapping, direct I/O, and cleanup out of the artifact parser. `artifact::Reader` should retain only `.ninfer`-specific range and alignment validation.

2. There is one remaining CUDA runtime linkage inconsistency. `tests/CMakeLists.txt` still links `ninfer_qwen3_6_35b_a3b_dflash_load_plan_test` directly against `CUDA::cudart`, while Windows selects `CUDA::cudart_static` transitively through `ninfer_core`.

This can place both the shared and static CUDA runtime targets in the final link graph. It may still build because the linker resolves symbols from whichever library appears first, but the effective runtime then becomes link-order-dependent and the test may unexpectedly acquire a `cudart64_*.dll` dependency.

Please route this target through the same project-level CUDA runtime selection as every other target. Ideally, define a single interface target:

```cmake
add_library(ninfer_cuda_runtime INTERFACE)

if(WIN32)
  set(CMAKE_CUDA_RUNTIME_LIBRARY Static)
  target_link_libraries(
    ninfer_cuda_runtime INTERFACE CUDA::cudart_static)
else()
  set(CMAKE_CUDA_RUNTIME_LIBRARY Shared)
  target_link_libraries(
    ninfer_cuda_runtime INTERFACE CUDA::cudart)
endif()
```

All consumers should then link only against `ninfer_cuda_runtime`, with no direct `CUDA::cudart` or `CUDA::cudart_static` references outside this definition.

PR #6, opened 2026-07-29
feat(kv-cache): support independent key and value formats

pinizhaninov:
## Summary
- Select K and V cache formats independently across the public API, layouts, runtime state, reporting, and tooling.
- Add mixed BF16/INT8-G64 prefill and decode kernels while retaining specialized symmetric paths.
- Extend CLI, serving, benchmarks, parity tooling, documentation, and real-artifact coverage for both mixed orders.

## Verification
- Windows CTest: 67/67 passed with the 27B and 35B-A3B artifacts.
- Windows pytest: 67 passed, 4 skipped because source checkpoint directories were unavailable.
- WSL CTest: 67/67 passed with the 27B and 35B-A3B artifacts.
- WSL pytest: 67 passed, 4 skipped because source checkpoint directories were unavailable.
- git diff --check passed.

### DIFF
--- a/apps/cli/main.cpp
+++ b/apps/cli/main.cpp
@@ -93,10 +93,6 @@ std::string format_finish(ninfer::FinishReason reason) {
     return "unknown";
 }
 
-std::string format_kv_cache(ninfer::KvCacheStorage storage) {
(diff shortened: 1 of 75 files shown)


Neroued:
Thanks for the PR, but I’d prefer to handle this kind of operator/kernel work myself.

PR #10, opened 2026-07-31
fix(serve): recover drifted Qwen tool calls

Sciguymjm:
## Summary

Qwen3.6 can occasionally drift from its XML-like tool-call format during long agentic sessions. NInfer's strict parser then returns the complete response as text, so clients such as OpenCode display a naked `<tool_call>` instead of receiving a structured tool call.

This adds an opt-in `--tolerant-tool-calls` serving mode that recovers complete function calls despite common Qwen wrapper/suffix drift, missing outer `</tool_call>`, or a call emitted before `</think>`.

## Design

- Strict parsing remains the default.
- Tolerant recovery only promotes a complete function with parsed parameters; incomplete calls are not executed.
- Streaming tolerant tool requests buffer reasoning until parsing completes, preventing malformed call text from leaking as reasoning or being emitted twice.
- Request-log schema is bumped to v5 and the benchmark consumer is updated accordingly.

## Verification

- Built the serving binary and focused C++ tests.
- `ninfer_tool_call_parser_test`: passed.
- `ninfer_serve_options_test`: passed.
- `ninfer_request_log_test`: passed.
- `PYTHONPATH=. python3 -m pytest -q tests/test_serve_corpus.py`: 2 passed.
- Manual validation was performed through OpenCode → llama-swap → NInfer against Qwen3.6-27B; the malformed/naked tool-call behavior was reproduced in the affected workflow.

No automated live-model HTTP integration fixture is included.


jmander11:
Have you ever seen this with qwen3.8-27b?

paolobarbolini:
I've seen it happen a few times a day. Very long thinking and/or writing long files seems to increase the likelyhood of it.

engharat:
I confirm this is happening to me too. I converted a QUASAR qwen3.8-27b NVFP4 model from vllm to ninfer and I am experiencing this tool broken command, and I was wondering where the issue was coming from.

gregtakacs:
Can confirm it is happening occasionally on 3.8-27b NVFP4 model.

gregtakacs:
Confirmed this still reproduces and this fix still resolves it as of today (2026-08-27), including on `qwen3_8_27b_nvfp4` (NVFP4) specifically — same symptom others reported here: occasional naked `<tool_call>` leaking as plain content under long agentic tool-use sessions (Cline, in our case).

Hand-merged this PR's actual fix onto a more current pin (`master`@9dbc0740, ~4 weeks and 68 commits ahead of where this branch forked) to unblock ourselves in the meantime, since `mergeable_state` here is `dirty` against current `master` (mainly request-log schema drift, v5→v17, unrelated to the actual fix). Built clean and all three relevant tests pass against the merged tree. Happy to share the resolved diff if it's useful for rebasing this branch — just say the word.

(Separately: filed #101 for an unrelated tool-call parsing bug we hit even with this fix applied — a strictly-typed parameter's JSON value containing an un-doubled backslash, e.g. a literal grep/regex \\* or \\|, rejects the whole call outright since your tolerant recovery here only engages once a call has already been parsed.)

Neroued:
NInfer is keeping tool call parsing strict as intended.

PR #13, opened 2026-08-11
feat(build): support native Windows builds

Chakratos:
Greetings!

some time ago i saw your post on reddit about ninfer, sadly it didn't had windows support.
Someone commented, that he simply told codex to make it windows compatible but he also changed it to support RTX 3090 instead of RTX 5090.
Since then i had my own branch with Windows compatability, for local testing.

Now i told codex to check my branch against your master and in order to prevent merge conflicts, he redid the windows support ontop of the current master branch.

Also Codex saw an already existing Windows PR (PR #4 ) , which did not satisfy your standards. The AI should have followed your findings, so that this branch should be clean :)

I hope everything is fine and you are as hyped as me about Qwen 3.8 30b tommorow!
Ninfer will be awesome for this i hope.

the following is ai generated (Codex 5.6 Sol High):

## Summary

- add a native Windows 11 build using Visual Studio, CUDA, and manifest-mode vcpkg
- move platform file access behind `core::ReadOnlyFile`, with separate POSIX and Win32 implementations
- route every project target through one `ninfer_cuda_runtime` interface target (static on Windows, shared on Linux)
- add the Win32 portability changes required by the CLI, server, artifact tools, and tests
- preserve the NVFP4 TMA kernels on Windows by placing tensor-map descriptors in stable device workspace storage
- document the supported Windows build and validation workflow

This supersedes #4 and directly incorporates its review feedback about file-access ownership and project-wide CUDA runtime selection.

## Why

The existing build was Linux-only at several independent boundaries: POSIX mapping/direct I/O lived inside the artifact reader, dependencies were discovered only through pkg-config, CUDA runtime linkage was selected ad hoc, and a few runtime/test call sites relied on POSIX or compiler-specific behavior. MSVC/NVCC also cannot use the existing by-value `__grid_constant__` tensor-map parameter route for these kernels.

The change keeps artifact validation in `artifact::Reader`, moves only OS file mechanics into `core`, centralizes CUDA runtime ownership, and uses the same public Engine path and model artifacts on both platforms.

## User impact

Windows users with an RTX 5090 can configure and build the CLI, HTTP server, and tests natively. Linux remains the default documented command-line build and retains the shared CUDA runtime and POSIX direct-I/O implementation.

## Validation

Validated on Windows 11, NVIDIA GeForce RTX 5090, CUDA 13.3, MSVC 19.51 / Visual Studio 2026, CMake 4.2:

- full Release build of the CLI, server, and tests
- CTest: 71 executed tests passed; 5 real-model tests skipped without their artifact environment variables
- artifact reader and artifact materialization tests rerun after the platform-file refactor
- real Qwen3.6-27B NVFP4 artifact load-plan path
- text generation through the native CLI with MTP and INT8 KV cache
- `py -3.11 -m py_compile tools/artifact/migrate_v1_to_v2.py`
- `git diff --check`

The available local WSL environment is too old for this tree (CMake 3.16 / CUDA 10.1), so a current Linux build was not rerun locally. The POSIX file implementation is an extraction of the existing reader implementation rather than a semantic replacement.


Neroued:
Thanks for the contribution. 
Unfortunately, native Windows support still requires platform-specific workarounds in core CUDA kernels, as shown by the NVFP4 changes in this PR. I don’t want to carry that additional complexity and maintenance burden. Since we already provide Docker support, please use Docker on Windows instead. For this reason, I won’t merge this PR.

Chakratos:
I understand, no problem :)
If i had saw the docker support, i wouldn't have made this PR.
But since the native windows version was made, I ran a comparison Benchmark.
I compared native Windows, Docker with the README bind mount, and Docker with a named volume on an RTX 5090:

- Qwen3.6-27B NVFP4
- MTP3, INT8 KV cache, CUDA Graph
- Same 18.3 GB `.ninfer` artifact and corpus
- Two runs per path, five measured repetitions per workload

| Runtime/storage path | Model load | PP512 tok/s | TG256 tok/s | PP512+TG256 prefill | PP512+TG256 decode |
| Native Windows | 4.507 s | 9,772.54 | 179.36 | 9,415.15 | 180.67 |
| Docker/WSL2 named volume | 5.554 s | 9,702.51 | 179.08 | 9,339.36 | 180.78 |
| Docker/WSL2 Windows bind mount | 85.083 s | 9,758.94 | 179.53 | 9,414.65 | 180.71 |

Inference performance is effectively identical. The improvement is storing the model in a Docker volume instead of loading it through Docker Desktop's Windows bind mount.

One-time setup:

​```powershell
docker volume create ninfer-models
docker run --rm -v ninfer-models:/models -v "${PWD}/models:/source:ro" alpine cp /source/qwen3_6_27b_nvfp4.ninfer /models/
​```

Then use `-v ninfer-models:/models:ro` when starting NInfer.

PR #24, opened 2026-08-15
feat(serve): advertise configured model context in models api

SolettaSolaris:
## Summary
- include `max_model_len` in native `/v1/models` and `/v1/models/{id}` entries
    - This is inline with what llama.cpp & vllm exposes in their models response 
- source the value from the same parsed `--max-context` configuration used for inference

### DIFF
--- a/docs/serving.md
+++ b/docs/serving.md
@@ -54,6 +54,39 @@ cannot be combined with `--vision`. A later request cannot enable a capability o
 | `POST /v1/messages` | Anthropic-style message generation |
 | `POST /v1/messages/count_tokens` | checkpoint-native expanded input-token count |
 
+### Models
+
+`GET /v1/models` returns a `list` object holding a single `data` entry, and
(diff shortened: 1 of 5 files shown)


matthewperron:
This would be useful for coding harness dynamic discovery of the context size.

gearwave00001:
This change seems to be necessary for hermes to correctly set the context length for the model.  I ran into an issue with hermes reporting the default max context of 131072 when ninfer is configured with 262144.  My hermes agent found the same issue as reported here and I had it write a patch, but I just saw this open ticket.  I'll pull this PR and see if it fixes my issue.

Edit: This fixed my issue and hermes works with ninfer now at max context.

SolettaSolaris:
max_model_len was added in https://github.com/Neroued/ninfer/commit/412b11f3e58cda66748b5da4d684ae76f2ccab64
Closing this as it's no longer needed.

Neroued:
implemented in 412b11f3

PR #35, opened 2026-08-17
feat(kv): compressed-KV cache (E8 lattice) for the Blackwell sm_120a path

danielfparkernz:
# PR: Live compressed-KV cache quantization on the Blackwell (sm_120a) path

**Branch:** `feat/compressed-kv-blackwell-5090`
**Base:** upstream `master` @ `32c9881` (or later)
**Scope:** compressed-KV cache (live KV-quantization) for the Qwen3.6 engine,
ported from the NVIDIA Ada (RTX 4090 / sm_89) fork onto the official
Blackwell (sm_120a / RTX 5090) codebase.

---

## What this adds

Four new live KV-quantization cache modes, selectable via `--kv-dtype`
(server and CLI):

| mode        | key storage                                   | value storage    | typical use      |
|-------------|-----------------------------------------------|------------------|------------------|
| `rk8v4`     | rotated int8 key, group-64 fp16 scale          | packed int4 V    | high-fidelity    |
| `rk4v4`     | rotated int4 key (packed)                      | packed int4 V    | long-context     |
| `rk4v4-e8`  | general Conway-Sloane **E8 lattice** key codes | packed int4 V    | long-context     |
| `rk2v4-e8`  | two-stage **E8 240-root** key codes, 2 bits/dim| packed int4 V    | max compression  |

The E8 modes are the headline: algebraic nearest-point projection onto the
Conway-Sloane E8 = D8 ∪ (D8 + ½·¹) lattice lets the K cache run at
**2–4 bits per dimension** instead of Full-Precision, so a 24 GB mobile GPU
can fit **262,144-token context** on a ~27B-parameter model — things a
bf16 or even a plain int8 KV cache can't hold at that length.

Commits:
1. `feat(ops)` — E8 lattice / root codecs, packed-rotation math, and the
   templated attention kernels, launchers, wrapper + data-model foundation.
2. `feat(serve)` — server `--kv-dtype` parsing + logging labels.
3. `feat(engine)` — runtime storage routing through the qwen3.6 planner.
4. `test(kv)` — standalone correctness oracle + build wiring + README.

## Why

24 GB-class Blackwell-class mobile GPUs (GeForce RTX 5090, sm_120a) can
serve long-context LLMs if the KV cache is compressed at attention time.
This brings the tried E8 compressed-KV machinery to the official Blackwell
tree so the market-facing architecture is the target-supported one.

## Attribution

Full credit goes to the original implementer:

- **UDPSendToFailed/ninfer-4090** — a fork of **Neroued/ninfer** target
  the NVIDIA Ada (sm_89 / GeForce RTX 4090) generation. **All of the
  compressed-KV cache design — the Hadamard-rotated K/V, the packed
  int4 V, and the E8 Conway-Sloane lattice / 240-root codec mathematics —
  originates here. Full credit to this fork's author.**
- itself derived from **Don-Chad/ninfer-3090**.

This branch is a **port of that source of truth onto the official
Blackwell (sm_120a) codebase**. The E8 codecs, the packing/rotation math,
and the registry of `rk*` modes are unchanged from the 4090 fork; the
changes here are the Blackwell target and the integration layer
(launchers, wrapper, runtime storage wiring). Attribution is also carried
in each commit body and in `tools/test_kv/README.md`.

## Verification evidence

Shipped with this PR is a standalone correctness oracle
(`tools/test_kv`, branch commit 4):

```text
NInfer: True E8 Conway-Sloane Lattice Microbenchmark
Corpus: 1,000,000 tokens, 256-dim, 5 embedded needles @ 5/25/50/75/95%

Method 1: Two-Stage E8 Root Codec (2-bit)  -> cosine ~100.0 vs FP32, needles PASSED
Method 2: General Conway-Sloane E8 Lattice -> cosine ~100.0 vs FP32, needles PASSED
Needle Retrieval Ranking: 5/5 [PASSED - 100% RETRIEVED]
```

- **100% needle retrieval** across 1M tokens for both codecs — the needles
  (high-magnitude directional key/query targets) are recovered at their
  exact indices by the quantized attention returning the same top scores
  as the full-FP32 reference.
- On this hardware an end-to-end sanity run of `rk2v4-e8` served a
  Qwen3.6-27B NVFP4 model **generating correctly at 262,144-token
  context** on a 24 GB-class Blackwell-class GPU.

### Reproduce

```sh
nvcc -arch=sm_120a -O3 tools/test_kv/test_e8_codec.cu \
     tools/test_kv/verify_1m_retrieval.cu -o verify_1m_retrieval
./verify_1m_retrieval            # 1,000,000 tokens default
./verify_1m_retrieval 2000000    # larger corpus, if VRAM allows
```

(or `cmake --build build --target ninfer_kv_e8_verify`; see
`tools/test_kv/README.md`).

## Scope / non-goals

- Adds the **E8 codec + compressed-KV storage** to the Qwen3.6 engine
  path. No changes to weights, no new formats, no changes to the
  Full-Precision path (defaults unchanged).
- Only the Blackwell (sm_120a) architecture is supported by NInfer and by
  this port (the reuse is source-level; there is no sm_89 target here).
- Building/running the oracle and the engine are unchanged otherwise.

### DIFF
--- a/apps/cli/options.cpp
+++ b/apps/cli/options.cpp
@@ -55,6 +55,10 @@ float parse_float(const char* text, std::string_view label, float minimum, float
 KvCacheStorage parse_kv_cache(std::string_view text) {
     if (text == "bf16") { return KvCacheStorage::BFloat16; }
     if (text == "int8") { return KvCacheStorage::Int8Group64; }
+    if (text == "rk8v4") { return KvCacheStorage::RotatedInt8KeyInt4ValueGroup64; }
+    if (text == "rk4v4") { return KvCacheStorage::RotatedInt4KeyInt4ValueGroup64; }
+    if (text == "rk4v4-e8") { return KvCacheStorage::RK4V4E8; }
+    if (text == "rk2v4-e8") { return KvCacheStorage::RK2V4E8; }
     throw std::invalid_argument("invalid kv-dtype: " + std::string(text));
 }
 
@@ -75,9 +79,9 @@ ReasoningEffort parse_reasoning_effort(std::string_view text) {
 std::string usage_text(const char* argv0) {
     return std::string("usage: ") + argv0 +
            " <model.ninfer> (--prompt <text>|--messages <messages.json>)\n"
-           "       [--max-context N] [--kv-capacity N|auto] [--prefill-chunk N] [--max-new N]\n"
-           "       [--device N]\n"
-           "       [--kv-dtype bf16|int8] [--spec mtp|dflash --draft-tokens N]\n"
+           "[--max-context N] [--kv-capacity N|auto] [--prefill-chunk N] [--max-new N]\n"
+           "[--device N]\n"
+           "[--kv-dtype bf16|int8|rk8v4|rk4v4|rk4v4-e8|rk2v4-e8] [--spec mtp|dflash --draft-tokens N]\n"
            "       [--lm-head-draft]\n"
            "       [--temperature F] [--top-p F] [--top-k N] [--min-p F]\n"
            "       [--presence-penalty F] [--frequency-penalty F] [--seed N] [--greedy]\n"
--- a/docs/pr-compressed-kv-blackwell.md
+++ b/docs/pr-compressed-kv-blackwell.md
@@ -1,0 +1,102 @@
+# PR: Live compressed-KV cache quantization on the Blackwell (sm_120a) path
+
+**Branch:** `feat/compressed-kv-blackwell-5090`
+**Base:** upstream `master` @ `32c9881` (or later)
+**Scope:** compressed-KV cache (live KV-quantization) for the Qwen3.6 engine,
+ported from the NVIDIA Ada (RTX 4090 / sm_89) fork onto the official
+Blackwell (sm_120a / RTX 5090) codebase.
+
+---
+
+## What this adds
+
+Four new live KV-quantization cache modes, selectable via `--kv-dtype`
+(server and CLI):
+
+| mode        | key storage                                   | value storage    | typical use      |
+|-------------|-----------------------------------------------|------------------|------------------|
+| `rk8v4`     | rotated int8 key, group-64 fp16 scale          | packed int4 V    | high-fidelity    |
+| `rk4v4`     | rotated int4 key (packed)                      | packed int4 V    | long-context     |
+| `rk4v4-e8`  | general Conway-Sloane **E8 lattice** key codes | packed int4 V    | long-context     |
+| `rk2v4-e8`  | two-stage **E8 240-root** key codes, 2 bits/dim| packed int4 V    | max compression  |
+
+The E8 modes are the headline: algebraic nearest-point projection onto the
+Conway-Sloane E8 = D8 ∪ (D8 + ½·¹) lattice lets the K cache run at
+**2–4 bits per dimension** instead of Full-Precision, so a 24 GB mobile GPU
+can fit **262,144-token context** on a ~27B-parameter model — things a
+bf16 or even a plain int8 KV cache can't hold at that length.
+
+Commits:
+1. `feat(ops)` — E8 lattice / root codecs, packed-rotation math, and the
+   templated attention kernels, launchers, wrapper + data-model foundation.
+2. `feat(serve)` — server `--kv-dtype` parsing + logging labels.
+3. `feat(engine)` — runtime storage routing through the qwen3.6 planner.
+4. `test(kv)` — standalone correctness oracle + build wiring + README.
+
+## Why
+
+24 GB-class Blackwell-class mobile GPUs (GeForce RTX 5090, sm_120a) can
+serve long-context LLMs if the KV cache is compressed at attention time.
+This brings the tried E8 compressed-KV machinery to the official Blackwell
+tree so the market-facing architecture is the target-supported one.
+
+## Attribution
+
+Full credit goes to the original implementer:
+
+- **UDPSendToFailed/ninfer-4090** — a fork of **Neroued/ninfer** target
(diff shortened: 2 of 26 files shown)


danielfparkernz:
Thanks for the detailed automated review — all three findings are legitimate and now addressed in commit 74ef824 (pushed to this branch).

1. **P1 — subgroup divergence in e8_root_codec.cuh**: Fixed. The early-return on rad_idx==0 (which could leave some 8-lane subgroups out of full-warp __shfl*_sync calls) is removed. All lanes now converge on every shuffle; a zero-rad_idx subgroup participates in the shuffles and emits the zero code via a guarded write after convergence. Reasoning is documented in the code comment.

2. **P2 — half-integer E8 coset collapse**: Documented as a deliberate half-coset approximation. e8_project_8d_warp can select a D8+0.5 point, and since the i4/int8 packed codes carry no coset bit, the half is intentionally collapsed by rintf()+cast and not reconstructed on decode. A clear comment now states this is an approximation (not an exact E8 lattice projection) for half-integral inputs; the rk4v4 non-E8 path is unaffected.

3. **P2 — verifier can never fail**: Fixed. verify_1m_retrieval now returns EXIT_FAILURE unless every needle is recovered at its exact token index, so CI/CTest can detect codec regressions.

Rebuilt for sm_120a (clean) and the codec still passes the 1M-token retrieval oracle. Design credit for the ported E8 codec remains with UDPSendToFailed/ninfer-4090 and the Don-Chad/ninfer-3090 lineage.

tiequan12345:
Validated this branch end-to-end on **Qwen3.8-27B NVFP4** (RTX 5090 32 GB, native CUDA build, MTP k=3, vision on). Adding the measured numbers in case they help the merge decision.

**Fit at 262,144 ctx — the interesting part.** The NVFP4 artifact (20.02 GiB weights) cannot reach 262k with INT8 KV: the runtime reservation is 13.02 GiB against 11.66 GiB available after weights. With `rk8v4` it fits — but only with an **explicit** `--kv-capacity 262144`. Auto capacity adds its 1 GiB headroom (10,740,308,992 + 1,073,741,824 bytes vs 11,657,122,304 available → ~150 MiB short) and the boot fails. Worth a line in the docs if `auto` stays default.

**Result: boots at 31.15 GiB, 262,144 ctx + MTP k=3 + vision — first configuration on this card that gets full context and the NVFP4 prefill path at the same time.**

**Retrieval quality at range (3-needle, distinct codes):** pass at 128,163 / 192,185 / 240,193 / 260,202 prompt tokens; identical 260k prefix re-use 0.31–0.33 s. No degradation we can detect vs INT8 KV.

**Task quality (paired A/B on a 135-request multimodal grading battery, two passes):** MAE 0.663 / 0.633; paired vs the same serving stack on groupwise-int weights + INT8 KV @ 262k: Δ −0.011, bootstrap 95% CI [−0.089, +0.067] — statistical tie. Decode ~180 tok/s (spec acceptance ~2.7); prefill 7.7k tok/s short-context.

**Long-context speed:** 128k end-to-end 38.5 s, 260k 133.6 s (groupwise-int/INT8 baseline: 58.1 s / 163.4 s) — 1.2–1.5× faster.

Did not exercise `rk4v4` / `rk2v4-e8` (rk8v4 already recovered the memory needed on 32 GB). Thanks for the port — this is now our production default.

danielfparkernz:
Rebased onto current upstream master (`feaf4dd`, Aug 19 18:36 UTC). No conflicts — the 6 newer upstream commits are qwen3.8 eval/benchmark additions that don't touch the compressed-KV patch files. Verified the rebased patch touches the identical 26-file set (+2717/-125) as before, so no content changed, only new SHAs. GitHub reports mergeable=true (cleanly merges into master).

This PR is now rebase-ready / review-ready. The E8 compressed-KV port (rk2v4-e8 / rk4v4 / rk4v4-e8 / rk8v4) enables 262K-token context on 24 GB Blackwell-class GPUs (sm_120a), verified end-to-end: 100% needle retrieval over a 1M-token corpus by both codecs, and Qwen3.6-27B NVFP4 + Qwen3.8-27B generating correctly at full 262,144-token context.

DuncanBetts:
I haven't smoke tested this nor reviewed the code but I've added it to my vibe coded fork of ninfer and with 25.31GB VRAM am running ninfer with dtype rk4v4-e8 concurrency 4 mtp 4 ctx 262144. Decode seems to be about 40%-50% slower than int8 on a RTX 5090 32GB (power limited to 400W).

Given the slowdown, and accuracy issues mentioned here [Don-Chad/ninfer-3090 README line 103-133](https://github.com/Don-Chad/ninfer-3090/blob/release/v0.6.0-rtx3090/README.md?plain=1#L103-L133) I've switched back to int8 for general work, but can see it would be useful if I need to run many short sessions concurrently; quantisation errors don't have the same opportunity to compound in such scenarios.

cometkim:
Hey guys, I have another experiment to support 1M context using the nvfp4 model + [HyperQuant](https://arxiv.org/abs/2606.23406) KV (3.7x compressed) + YaRN. In my [fork](https://github.com/cometkim/ninfer). I think this is somehow newer research that shares a similar base idea with this PR.

I wanted to let you know that this quantization method works quite well. The whole context fits on my single 5090 (even 1M if I disabled vision or MTP), and up to 524k (2x scaling) is mostly fine for my use case.

I'm currently evaluating 2x RoPE scaling with hyperquant to unlock a 524k context window. Accuracy is ok-ish (no diff at GPQA diamond and AIME). Speed is about - 0.85x vs int8, but still optimizing.

1M needs x4 RoPE scaling, and it is noticeably getting dumb.

Neroued:
Superseded by 4ac73c47.

PR #43, opened 2026-08-18
Jinja template support - #42

Doelfke:
Qwen3.8-27b is amazing, but it's very chatty and still has compatibility issues. I have successfully tested these changes with https://huggingface.co/peculiar-ragdoll/Qwen-Sharp-Chat-Templates, a fork of the popular froggeric Qwen template fix.

### DIFF
--- a/apps/cli/main.cpp
+++ b/apps/cli/main.cpp
@@ -258,6 +258,7 @@ int main(int argc, char** argv) {
             std::cerr, ninfer::product::stderr_load_progress_options());
         ninfer::EngineOptions engine_options;
         engine_options.artifact_path  = cli.artifact_path;
+        engine_options.chat_template_path = cli.chat_template_path;
         engine_options.device         = cli.device;
         engine_options.max_context    = cli.max_context;
         engine_options.kv_capacity    = cli.kv_capacity;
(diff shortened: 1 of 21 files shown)


jmander11:
Do you think this is better than just adding a system prompt prepend option if you have something you wanted added generally and selectively fixing something like empty thinking tags injected with preserve thinking on and no reasoning content?

I don't think you want anything else from the froggeric chat template as they are not relevant to ninfer or are strictly worse like telling the model to fix arguments to a call that may have correctly errored out with successful arguments. That error fix seems targeted at small qwen 3.5 models.

https://github.com/Neroued/ninfer/commit/392b1877d9147acd4178488c4db1d9fd7647fcbc
https://github.com/Neroued/ninfer/commit/3a62d24021ffd10d9f99b7559167b84389f87a47

Doelfke:
> Do you think this is better than just adding a system prompt prepend option if you have something you wanted added generally and selectively fixing something like empty thinking tags injected with preserve thinking on and no reasoning content?
> 
> I don't think you want anything else from the froggeric chat template as they are not relevant to ninfer or are strictly worse like telling the model to fix arguments to a call that may have correctly errored out with successful arguments. That error fix seems targeted at small qwen 3.5 models.
> 
> [392b187](https://github.com/Neroued/ninfer/commit/392b1877d9147acd4178488c4db1d9fd7647fcbc) [3a62d24](https://github.com/Neroued/ninfer/commit/3a62d24021ffd10d9f99b7559167b84389f87a47)

That may be true, but these are the standard.

PR #48, opened 2026-08-18
fix(bench): accept request log schema v10

FlorianZimmer:
## Problem

`perf(vision): streamline multimodal preparation` raised the serving request-log schema from 9 to 10, but the serving corpus runner continued to require schema 9. As a result, the runner rejects a request log produced by `ninfer-serve` from the same revision before it can execute a benchmark.

## Change

- require request-log schema 10 in `run_serve_corpus.py`;
- update the regression to accept schema 10 and reject stale schema 9; and
- align the benchmark README with the current serving schema.

This does not change inference or the request-log format. It only makes the benchmark consumer match the current producer.

## Verification

```text
python -m pytest -q tests/test_serve_corpus.py
2 passed in 0.01s

python -m compileall -q \
  tools/bench/run_serve_corpus.py tests/test_serve_corpus.py

git diff --check HEAD^ HEAD
```

Before the constant was corrected, the focused regression reproduced the mismatch by rejecting a schema-10 `server_start` event while expecting schema 9.

I used OpenAI Codex (GPT-5) to diagnose the schema mismatch and implement this patch. I reviewed the complete diff and reran the focused verification above.


DuncanBetts:
Closed [https://github.com/Neroued/ninfer/pull/93](https://github.com/Neroued/ninfer/pull/93) as duplicate, though due to extreme PR message clarity, mine may make merging this easier. Both make identical changes.

PR #50, opened 2026-08-18
accept vLLM-style enable_thinking: top-level field and chat_template_kwargs

tiequan12345:
## Problem

Clients that target vLLM (and Qwen's own serving docs) control thinking mode with `chat_template_kwargs: {"enable_thinking": false}` inside `messages[0]` — vLLM's documented contract for Qwen3-family models. NInfer rejects this body shape, so every vLLM-targeted client needs a special-case branch before it can talk to NInfer. Some clients also send a plain top-level `enable_thinking`, which vLLM accepts as an alias.

## Change

- Accept optional top-level `enable_thinking` (bool) in `/v1/chat/completions` and the Responses API.
- Accept `chat_template_kwargs.enable_thinking` nested in the first user message, mapping it onto the same option (vLLM semantics).
- If both are present and disagree, reject with 400 `conflicting_template_option` (fail loudly rather than guess).
- All other `chat_template_kwargs` keys are ignored, and extra unknown keys inside `chat_template_kwargs` do not fail the request — same accept-and-ignore posture as #45.

Schema tests cover: nested-only, top-level-only, both agreeing, both conflicting, and `chat_template_kwargs` with unrelated keys.

## Context

Validated in production for two days on a Qwen3.8-27B grading service (RTX 5090, 262144 ctx, MTP): client code written against vLLM worked against NInfer unchanged once this landed, including bodies carrying both shapes. No behavior change for requests that omit both fields.

### DIFF
--- a/src/serve/openai_schema.cpp
+++ b/src/serve/openai_schema.cpp
@@ -479,6 +479,22 @@ std::string sse_event(const Json& payload) { return "data: " + payload.dump() + 
 
 } // namespace
 
+std::optional<bool> parse_openai_template_enable_thinking(const Json& body) {
+    if (!body.contains("chat_template_kwargs")) { return std::nullopt; }
+    const Json& kwargs = body.at("chat_template_kwargs");
+    if (!kwargs.is_object()) {
+        bad_request("chat_template_kwargs must be an object", "chat_template_kwargs");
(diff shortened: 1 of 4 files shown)


Neroued:
done in master

PR #55, opened 2026-08-19
feat(serve): report prefix-cache hits in the chat-completions usage

Sociopacific:
## Summary

Emit `usage.prompt_tokens_details.cached_tokens` on `/v1/chat/completions`, from the same `outcome.metrics.prefix_cache_hit_tokens` and with the same clamp the Responses API already uses for `input_tokens_details.cached_tokens`.

## Why this is worth merging

A client cannot see the Engine's prefix cache. The chat-completions usage block carried only `prompt_tokens` / `completion_tokens` / `total_tokens`, so any harness computing a cache-hit rate from the OpenAI protocol read a flat zero — on a session the Engine was in fact reusing at 92%, and 98.6% over its last fifty requests. Users conclude prefix reuse is broken and start disabling it or filing issues; the feature works, it is just invisible on the protocol most clients speak.

The number was already there. The request log has printed `prefix_cache_hit` tokens per request since it was written, and the Responses schema has reported `input_tokens_details.cached_tokens` from the start. This makes the two protocols agree instead of one of them silently reporting nothing.

`cached_tokens` is a subset of `prompt_tokens`, never an addend — that is what OpenAI documents and what clients subtract to get billed input, so existing consumers that ignore the field are unaffected.

## What changed

- `src/serve/request.h` — carry the hit count into the response-building path.
- `src/serve/openai_schema.cpp` — one shared helper builds the usage object. All three chat-completions call sites (plain, tool, and the streaming usage chunk) previously constructed the same object by hand; they now share it, which is why the diff removes more lines than a one-field addition would.
- `src/serve/http_server.cpp` — pass the metric through.

## Scope / non-goals

- The Anthropic Messages schema deliberately stays as it was. There `input_tokens` *excludes* cache reads, so reporting the same fact through it would change what `input_tokens` means for every existing Messages client.
- No change to how the prefix cache works, only to what is reported.

### DIFF
--- a/src/serve/http_server.cpp
+++ b/src/serve/http_server.cpp
@@ -398,7 +398,8 @@ void HttpServer::handle_chat_completions(const httplib::Request& req, httplib::R
                 return req.is_connection_alive && !req.is_connection_alive();
             });
             log_request_done(log_context, outcome);
-            const CompletionUsage usage{outcome.prompt_tokens, outcome.completion_tokens};
+            const CompletionUsage usage{outcome.prompt_tokens, outcome.completion_tokens,
+                                        static_cast<int>(outcome.metrics.prefix_cache_hit_tokens)};
             std::string response_body;
             if (!outcome.tool_calls.empty()) {
                 response_body = make_chat_completion_tool_response(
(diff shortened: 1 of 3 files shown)


Apoze:
I reproduced one follow-up issue while validating this PR on the Windows/RTX 5090 build: `tools/smoke/serve_contract.py` compares a cold non-streaming `usage` object with the repeated streaming request. The runtime correctly reports `cached_tokens` as `0` and then `38`, so the smoke test fails even though every other contract value matches.

A minimal test-only follow-up is available at Apoze/ninfer@cfbebe99692871de07250e8abc15f834a2f62cb1. It validates the cache count type and `0 <= cached_tokens <= prompt_tokens`, removes only that field from the cold/warm equality comparison, and keeps every other usage detail strict.

Verification: `py_compile`, focused invalid-value checks, `git diff --check`, and the complete `ninfer_serve_contract_v2` against the source-built Windows server all pass. I did not open a competing upstream PR because this commit depends on #55; please feel free to cherry-pick it.

Neroued:
superseded by eda40242

Issue #56, opened 2026-08-19
OpenAI Chat Completions: 400 on tool messages with array content


jtf9808:
## Symptom

Any agent client that serializes tool results as an **array of content parts** instead of a
bare string gets `400` on every tool call in an agentic loop. Qwen Code does this by default
(provider setting `toolResultContentFormat: "parts"`), so running it against ninfer's
`/v1/chat/completions` fails as soon as the first tool result is fed back.

The same shape is accepted for `user`/`assistant` messages, and both forms are part of the
advertised contract: the OpenAI spec defines tool `content` as string-or-array (only `text`
parts valid for tool messages), and `docs/serving.md` lists the array form.

## Environment

- ninfer `master` @ `a05746aa`
- Model: `qwen3.8:27B` (nvfp4)
- Client: Qwen Code (default `toolResultContentFormat: "parts"`)

## Reproduction

```bash
curl -s -w '\nHTTP_STATUS=%{http_code}\n' http://<host>:<port>/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "qwen3.8:27B",
    "messages": [
      {"role": "user", "content": "What is the weather in Paris?"},
      {"role": "assistant", "content": null, "tool_calls": [
        {"id": "call_1", "type": "function",
         "function": {"name": "get_weather", "arguments": "{\"city\":\"Paris\"}"}}]},
      {"role": "tool", "tool_call_id": "call_1",
       "content": [{"type": "text", "text": "{\"temp\":20,\"condition\":\"sunny\"}"}]}
    ]
  }'
```

**Actual:** `400`
```json
{
    "error": {
        "code": null,
        "message": "tool messages must contain string content",
        "param": "messages",
        "type": "invalid_request_error"
    }
}
```

**Expected:** `200` — the array form is what Qwen Code and other clients send by default; it
should be accepted the same way it already is for `user`/`assistant` messages.

## Root cause

`parse_messages` (`src/serve/openai_schema.cpp`, lines 253-258 at `a05746aa`) hard-requires a
string for tool content, instead of routing it through `parse_content_parts()` like the other
roles.

## Workaround (client side)

In Qwen Code set the provider's `generationConfig.toolResultContentFormat` to `"string"` so tool results are
sent as a bare string.

Neroued:
fixed in 79c292bc07e5edb349cb9311136c3aad8631d80c

PR #57, opened 2026-08-19
Fix/tool message content parts

jtf9808:
Problem

`POST /v1/chat/completions` rejects tool messages whose `content` is a content-part array
(`[{ "type": "text", "text": ... }]`) with `400 "tool messages must contain string content"`.
OpenAI-compatible agent clients emit that shape by default (Qwen Code serializes tool results
as parts), so every tool call in an agentic loop fails. The OpenAI spec defines tool `content`
as oneOf string | content-part array (`minItems: 1`), only `type: "text"` parts valid; other
roles already accept both forms via `parse_content_parts()`, but the tool branch
(`openai_schema.cpp`) hard-required the string.

## Change

Route tool content through the existing `parse_content_parts()` helper with a new optional
`allowed_types` allowlist (wire type names, default empty = accept all, behavior unchanged):

- `parse_content_parts(content, turn, index, allowed_types = {})` — when non-empty, a guard in
  the part loop rejects any part whose `type` is not in the list with a precise 400.
- Tool branch: after its existing `tool_call_id`/`content` checks, calls
  `parse_content_parts(item.at("content"), turn, i, {"text"})`.
- All other call sites omit the argument — user/assistant/system behavior unchanged.
- `parse_content_parts()` is exported in `openai_schema.h` (moved out of the anonymous
  namespace, same as `parse_openai_preserve_thinking`) so the allowlist is unit-testable.

Non-text parts (media, audio, unknown) are rejected at the schema boundary with a precise 400
instead of being media-acquired and failing later with `modality_not_supported`.

## Affected behavior

| Input | Before | After |
|---|---|---|
| `content` = string | accepted | accepted (unchanged) |
| `content` = `[{"type":"text","text":...}]` | **400** | accepted, parsed to text parts |
| multiple text parts | **400** | accepted, kept separate (matches other roles) |
| non-text part (e.g. `image_url`) | **400** | 400 `"content parts must have type 'text'"` |
| `content` = `[]` | 400 | 400 `"content must not be empty"` |
| `content` = null/missing | 400 | 400 `"tool messages must contain content"` |

Downstream translation, prompt construction, sampling, and non-tool roles are unaffected.
`docs/serving.md` documents the tool-message content contract in the same change.

## Verification

Base `a05746aa`, tip `f197be4b` (RTX 5090, `sm_120a`, CUDA 13.1).

- `ninfer_openai_schema_test` passed — new `test_parse_tool_message_content_parts` (array
  accepted, multiple text parts kept separate, bare string still accepted, empty array/null
  rejected, `image_url`/`video_url`/`input_audio`/unknown rejected at the schema boundary) and
  new `test_parse_content_parts_allowed_types` (default allowlist accepts all normal part
  types; multi-type allowlist accepts listed types and rejects the rest, listing allowed types
  in caller order; guard runs before any media acquisition; empty array still rejected).
- Serve-protocol suite: **4/4 passed** (openai/anthropic/responses schema, tool-call parser) —
  no cross-protocol regression in the shared `ninfer_serve` layer.
- Full CTest (89 tests): 79 passed, 10 failed, 5 skipped; all 10 failures are environmental
  (9 × GPU OOM from a live serve process, 1 × hardcoded model paths) and none link
  `ninfer_serve`. All 11 serve/protocol contract tests pass.
- `git diff --check` clean.
- End-to-end (live server, `qwen3.8:27B` nvfp4): the reproduction payload (tool result as
  text-part array) returns **HTTP 200** with a valid completion; the identical payload on the
  unpatched server returns 400.

## AI-assisted implementation disclosure

Implementation and tests drafted in an AI-assisted session (Qwen Code agent; model
`qwen3.8:27B`) and reviewed by the contributor.

#56

### DIFF
--- a/docs/serving.md
+++ b/docs/serving.md
@@ -73,6 +73,8 @@ The endpoint supports:
 
 - `system`, `developer`, `user`, `assistant`, and `tool` history;
 - string content and ordered text, `image_url`, and `video_url` parts;
+- tool messages with string content or an array of `text` content parts (the OpenAI
+  contract allows no other part type for tool messages);
 - `max_completion_tokens` and the legacy `max_tokens` spelling;
 - `temperature`, `top_p`, `top_k`, presence/frequency penalties, and a nonnegative `seed`;
 - one stop string or an array of stop strings;
--- a/src/serve/openai_schema.cpp
+++ b/src/serve/openai_schema.cpp
@@ -137,60 +137,6 @@ ninfer::product::media_acquire::Source parse_media_url(const Json& part, const c
     return source;
 }
 
-void parse_content_parts(const Json& content, ChatTurn& turn, std::size_t index,
-                         std::vector<std::string> allowed_types = {}) {
-    if (content.is_string()) {
-        turn.content.push_back(ContentPart{ContentKind::Text, content.get<std::string>(), "text"});
-        return;
-    }
-    if (!content.is_array()) {
-        bad_request("message " + std::to_string(index) + " content must be a string or array",
-                    "messages");
-    }
-    std::string allowed_list;
-    for (const std::string& allowed : allowed_types) {
-        if (!allowed_list.empty()) { allowed_list += ", "; }
-        allowed_list += "'" + allowed + "'";
-    }
-    for (const Json& part : content) {
-        if (!part.is_object() || !part.contains("type") || !part.at("type").is_string()) {
-            bad_request("message " + std::to_string(index) +
-                            " content parts must be objects with a string 'type'",
-                        "messages");
-        }
-        const std::string type = part.at("type").get<std::string>();
-        if (!allowed_types.empty() &&
-            std::find(allowed_types.begin(), allowed_types.end(), type) == allowed_types.end()) {
-            bad_request("message " + std::to_string(index) + " content parts must have type " +
-                            allowed_list,
-                        "messages");
-        }
-        ContentPart out;
-        out.type_raw = type;
-        if (type == "text") {
-            if (!part.contains("text") || !part.at("text").is_string()) {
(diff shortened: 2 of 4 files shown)


DuncanBetts:
> Problem
> 
> `POST /v1/chat/completions` rejects tool messages whose `content` is a content-part array (`[{ "type": "text", "text": ... }]`) with `400 "tool messages must contain string content"`.

I can confirm I received this error when using Qwen Coder, so I'm using pi instead.

I'm not a C++ developer so my only input might be to consider restoring `parse_content_parts` to its original position in `src/serve/openai_schema.cpp` to simplify diffing/review.

jtf9808:
Fair point on the move, it is a bit annoying in the diff. I moved it because the new unit test can't reach it from the anonymous namespace so it now sits alongside the other exported helpers.

devan-carlin:
Thanks for this — the allowlist design and the precise 400s are a nice touch.

One gap I hit in practice: VS Code Copilot Chat's browser-screenshot tool returns the captured image as an `image_url` content part inside the tool result (not a text part). With the `{"text"}` allowlist, that request still 400s (`content parts must have type 'text'`), so the screenshot-only agentic flow against a vision model remains broken.

I verified the permissive variant end-to-end: routing tool content through `parse_content_parts()` with no allowlist (same as user/assistant) makes the Copilot Chat screenshot turn complete against a local vision model, while plain-string tool results are unchanged. Branch: `devan-carlin:fix/tool-message-content-parts` (single commit, 6 lines in `openai_schema.cpp`).

Happy to fold that in here if you'd like the allowlist to default to accepting media parts for tool messages (or gate it on the model's vision capability), rather than landing as a separate PR.

PR #59, opened 2026-08-19
Windows support: build and run NInfer on Windows 11 (MSVC + CUDA + Ninja)

pelebel:
## Summary

Build and run `ninfer` / `ninfer-serve` on **Windows 11** with MSVC (VS2022 `cl`) + `nvcc` (CUDA 13.x) + Ninja, resolving FFmpeg and libcurl from a vcpkg `x64-windows` prefix. Both executables build and run on an RTX 5090 (`sm_120a`); all CRT / CUDA / FFmpeg / curl / vcpkg runtime DLLs resolve in place.

This opens the engine (currently Linux/POSIX-only) to Windows dev/eval environments.

## Key changes

- **CMake**
  - Per-language MSVC flags (`/Zc:preprocessor`, `NOMINMAX`, `UTF8PROC_STATIC`) applied to C/CXX only so CUDA object files keep their build cache (`CMakeLists.txt`).
  - New `cmake/NInferMediaDeps.cmake` locates FFmpeg/curl — pkg-config on POSIX, vcpkg prefix on Windows — and exposes `NInfer::FFmpeg` / `NInfer::Curl`. `apps/CMakeLists.txt` copies the runtime DLLs next to the executables at build time (Windows only). `Ws2_32` is linked for the socket-using libs (Windows only).
  - The Windows guards no longer mis-fire on clang-cl. CMake sets `MSVC` for any MSVC-ABI compiler, so `/Zc:preprocessor` (a `cl` flag that clang-cl reports as an unknown argument) is now gated on the C++ frontend really being `cl`; the `-Xcompiler` copy stays unconditional because nvcc drives `cl` as its host compiler regardless. See the clang-cl note below.
- **Portability shims** (every Windows construct is behind `#ifdef _WIN32` with the original POSIX path in `#else`)
  - `src/artifact/reader.cpp`: `CreateFileW` / `MapViewOfFile` mapped-file reader (vs `open` / `fstat` / `mmap`).
  - `src/product/media_acquire/acquire.cpp`: Winsock (`WSAStartup`, `getaddrinfo`, `inet_ntop`).
  - `src/serve/console_log.cpp` (`localtime_s`), `src/serve/request_log.cpp` (`GetCurrentProcessId`), `src/product/load_progress/load_progress.cpp` (`_isatty`).
- **Artifact reader fixes** (both Windows-only, both found after the first revision of this PR)
  - The size check cast `size_t`'s maximum to `LONGLONG` before comparing. That cast wraps to `-1` on a 64-bit build, so *every* artifact compared greater and was rejected with "artifact size does not fit the process address space" before the mapping was attempted. Now compared in `uintmax_t`, matching the POSIX path directly below it. Reported by @AstrandPallas — thank you.
  - The mapped direct read did not mirror `pread`'s short-read semantics at end-of-file.
- **NVFP4 TMA** (`src/ops/linear/nvfp4/*`, `src/ops/linear_swiglu/nvfp4/*`): pass the 128-byte descriptors by device pointer so `cl` accepts the aligned by-value parameters. Cross-platform-safe (valid CUDA C++ under `nvcc` on Linux too).
- **qwen3_6 plan move ctors** (`src/targets/qwen3_6/impl/runtime/api_impl.h`): give the `SequencePlan` / `RequestBasePlan` / `RequestPlan` move constructors explicit bodies. MSVC does not emit an out-of-line `= default` explicit specialization unless it is ODR-used, which left these unresolved at link time; the explicit bodies are behavior-identical and also compile under GCC/Clang.
- **Test suite under MSVC** (`tests/`): the suite now configures, builds, and runs on Windows. Tests that need FFmpeg/curl DLLs get those directories prepended to `PATH` per-test rather than duplicating DLLs into every output directory.
- **CI**: `.github/workflows/windows-build.yml` adds a build-only job on a hosted `windows-2022` runner. Hosted runners have no GPU, so `ctest` is not executed there; the job guards the Windows *compile* path and smoke-launches both executables.

## Linux is not affected

Every Windows-specific construct is guarded (`if(MSVC)` / `if(WIN32)` / `#ifdef _WIN32` with a POSIX `#else`), and the POSIX paths are the original, unchanged code. FFmpeg/curl still resolve via pkg-config on POSIX. The only shared-source changes (TMA by-pointer, explicit move-ctor bodies) are behavior-preserving and compile on the Linux toolchain.

## Build on Windows

```bat
call "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvars64.bat"
vcpkg install curl ffmpeg --triplet x64-windows   # if not already installed
cmake -S . -B build-win -G Ninja -DCMAKE_BUILD_TYPE=Release ^
      -DCMAKE_CUDA_ARCHITECTURES=120a ^
      -DNINFER_MEDIA_ROOT=<vcpkg-root>/installed/x64-windows
cmake --build build-win -j
```

Requires CUDA >= 13.1 and exactly `sm_120a` (RTX 5090), same as the existing platform contract.

## Verification status

Verified on Windows 11 (RTX 5090, CUDA 13.3.73, VS2022 `cl` 14.44, Ninja, vcpkg `x64-windows`), on this exact branch:

- **Clean configure and build from scratch** — exit 0, both `ninfer.exe` and `ninfer-serve.exe` produced.
- **Full test suite: 89/89 passed** under `ctest`, including the numerical Op suites (NVFP4 / FP8 / W8A16 linear and SwiGLU) on real hardware. Six tests self-skip (return code 77) because they need a local Qwen3.6-27B HF checkout or a second artifact that is not on this machine.
- **Real GPU inference** — `ninfer-serve` loads a 17 GB Qwen3.8-27B artifact and answers `/v1/chat/completions` correctly. Sample single-request figures, no speculative decoding: `prompt=67 gen=47 ttft=104ms prefill=647.7tok/s decode=74.4tok/s`. (Short generation; indicative, not a benchmark.)

Remaining caveats, stated plainly:

- **Not compiled on Linux in this PR.** The Windows CI job above covers the Windows compile path; confirming the POSIX build stays green is still worth a maintainer's eye or a Linux job.
- **Not tested with clang-cl.** clang-cl is not installed on my machine, so I make no claim that it works. Note that nvcc on Windows only supports `cl.exe` as its host compiler, so even a clang-cl build would route the CUDA TUs through nvcc -> cl. The CMake guards have been corrected so an attempt starts from a sane place (see @AmerM137's comment thread).
- **No Linux-vs-Windows performance comparison.** No Linux or WSL install on the same box; I would rather not quote a number I cannot stand behind.

### DIFF
--- a/.github/workflows/qwen-dispatch.yml
+++ b/.github/workflows/qwen-dispatch.yml
@@ -1,0 +1,204 @@
+name: '🔀 Qwen Code Dispatch'
+
+on:
+  pull_request_review_comment:
+    types:
+      - 'created'
+  pull_request_review:
+    types:
+      - 'submitted'
+  pull_request:
+    types:
+      - 'opened'
+  issues:
+    types:
+      - 'opened'
+      - 'reopened'
+  issue_comment:
+    types:
+      - 'created'
+
+defaults:
+  run:
+    shell: 'bash'
+
+jobs:
+  debugger:
+    if: |-
+     ${{ fromJSON(vars.DEBUG || vars.ACTIONS_STEP_DEBUG || false) }}
+    runs-on: 'ubuntu-latest'
+    permissions:
+      contents: 'read'
+    steps:
+      - name: 'Print context for debugging'
+        env:
+          DEBUG_event_name: '${{ github.event_name }}'
+          DEBUG_event__action: '${{ github.event.action }}'
+          DEBUG_event__comment__author_association: '${{ github.event.comment.author_association }}'
+          DEBUG_event__issue__author_association: '${{ github.event.issue.author_association }}'
+          DEBUG_event__pull_request__author_association: '${{ github.event.pull_request.author_association }}'
+          DEBUG_event__review__author_association: '${{ github.event.review.author_association }}'
+          DEBUG_event: '${{ toJSON(github.event) }}'
+        run: |-
+          env | grep '^DEBUG_'
+
+  dispatch:
+    # For PRs: only if not from a fork
+    # For issues: only on open/reopen
+    # For comments: only if user types @gemini-cli and is OWNER/MEMBER/COLLABORATOR
+    if: |-
+      (
+        github.event_name == 'pull_request' &&
+        github.event.pull_request.head.repo.fork == false
+      ) || (
+        github.event_name == 'issues' &&
+        contains(fromJSON('["opened", "reopened"]'), github.event.action)
+      ) || (
+        github.event.sender.type == 'User' &&
+        startsWith(github.event.comment.body || github.event.review.body || github.event.issue.body, '@qwen-code') &&
+        contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association || github.event.review.author_association || github.event.issue.author_association)
+      )
+    runs-on: 'ubuntu-latest'
+    permissions:
+      contents: 'read'
+      issues: 'write'
+      pull-requests: 'write'
+    outputs:
+      command: '${{ steps.extract_command.outputs.command }}'
+      request: '${{ steps.extract_command.outputs.request }}'
+      additional_context: '${{ steps.extract_command.outputs.additional_context }}'
+      issue_number: '${{ github.event.pull_request.number || github.event.issue.number }}'
+    steps:
+      - name: 'Mint identity token'
+        id: 'mint_identity_token'
+        if: |-
+          ${{ vars.APP_ID }}
+        uses: 'actions/create-github-app-token@a8d616148505b5069dccd32f177bb87d7f39123b' # ratchet:actions/create-github-app-token@v2
+        with:
+          app-id: '${{ vars.APP_ID }}'
+          private-key: '${{ secrets.APP_PRIVATE_KEY }}'
+          permission-contents: 'read'
+          permission-issues: 'write'
+          permission-pull-requests: 'write'
+
+      - name: 'Extract command'
+        id: 'extract_command'
+        uses: 'actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea' # ratchet:actions/github-script@v7
+        env:
+          EVENT_TYPE: '${{ github.event_name }}.${{ github.event.action }}'
+          REQUEST: '${{ github.event.comment.body || github.event.review.body || github.event.issue.body }}'
+        with:
+          script: |
+            const eventType = process.env.EVENT_TYPE;
+            const request = process.env.REQUEST;
+            core.setOutput('request', request);
+
+            if (eventType === 'pull_request.opened') {
+              core.setOutput('command', 'review');
+            } else if (['issues.opened', 'issues.reopened'].includes(eventType)) {
+              core.setOutput('command', 'triage');
+            } else if (request.startsWith("@qwen-code /review")) {
+              core.setOutput('command', 'review');
+              const additionalContext = request.replace(/^@qwen-code \/review/, '').trim();
+              core.setOutput('additional_context', additionalContext);
+            } else if (request.startsWith("@qwen-code /triage")) {
+              core.setOutput('command', 'triage');
+            } else if (request.startsWith("@qwen-code")) {
+              const additionalContext = request.replace(/^@qwen-code/, '').trim();
+              core.setOutput('command', 'invoke');
+              core.setOutput('additional_context', additionalContext);
+            } else {
+              core.setOutput('command', 'fallthrough');
+            }
+
+      - name: 'Acknowledge request'
+        env:
+          GITHUB_TOKEN: '${{ steps.mint_identity_token.outputs.token || secrets.GITHUB_TOKEN || github.token }}'
+          ISSUE_NUMBER: '${{ github.event.pull_request.number || github.event.issue.number }}'
+          MESSAGE: |-
+            🤖 Hi @${{ github.actor }}, I've received your request, and I'm working on it now! You can track my progress [in the logs](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}) for more details.
+          REPOSITORY: '${{ github.repository }}'
+        run: |-
+          gh issue comment "${ISSUE_NUMBER}" \
+            --body "${MESSAGE}" \
+            --repo "${REPOSITORY}"
+
+  review:
+    needs: 'dispatch'
+    if: |-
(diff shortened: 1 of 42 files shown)


AmerM137:
It would be good to test the build with clang-cl also as it's a common compiler choice for a lot of software on windows. It's my preferred way to compile llama.cpp on a windows machine.

aresstokrat:
wow! waiting for this...did you compare speed to WSL or native Linux version?

AstrandPallas:
hit a bug building this on windows, the artifact size check casts SIZE_MAX to a signed LONGLONG so it wraps to -1 and rejects every model with 'artifact size does not fit the process address space'. switched it to uint64 and it loads and runs fine now.

Skiller87:
> hit a bug building this on windows, the artifact size check casts SIZE_MAX to a signed LONGLONG so it wraps to -1 and rejects every model with 'artifact size does not fit the process address space'. switched it to uint64 and it loads and runs fine now.

@AstrandPallas 

Could you please tell me how you did it? I’m getting the same error and haven’t found a way to fix it.

pelebel:
@AstrandPallas Confirmed, and thank you for the precise diagnosis — that is exactly it. `SIZE_MAX` cast to `LONGLONG` is `-1` on a 64-bit build, so every artifact of size >= 0 compared greater and was rejected before the mapping was ever attempted.

Fixed in `1fecf28e` by comparing in `uintmax_t`, which also makes the Windows path match the POSIX path directly below it:

```cpp
if (size_info.QuadPart < 0 ||
    static_cast<std::uintmax_t>(size_info.QuadPart) >
        std::numeric_limits<std::size_t>::max()) {
```

The updated branch loads a 16.7 GiB artifact and serves requests on an RTX 5090.

pelebel:
@AmerM137 Reasonable ask, and I do not want to claim more than I have tested: clang-cl is not installed on my machine, so this PR is **not** clang-cl-verified.

Two things worth flagging:

1. nvcc on Windows only supports `cl.exe` as its host compiler, so even a clang-cl build routes the CUDA translation units through nvcc -> cl. Only the C/C++ TUs would actually be clang-cl.
2. While looking into it I found the build was quietly mis-flagging clang-cl. CMake sets `MSVC` for any MSVC-ABI compiler, clang-cl included, so it was being handed `/Zc:preprocessor` (which clang-cl reports as an unknown argument on every TU), and the op tests' `-fno-fast-math` / `-ffp-contract=off` were being passed in a spelling clang-cl ignores — which would have dropped that guarantee on the CPU reference the GPU kernels are checked against, with no diagnostic.

Both are fixed in `9028b8fa` so a clang-cl attempt at least starts from a sane place. If you try it, I would be glad to fold in whatever else it needs.

pelebel:
@aresstokrat No controlled comparison yet — I do not have Linux or WSL on this box, so I would rather not quote a speedup I cannot stand behind.

For raw reference, the native Windows build on an RTX 5090 with Qwen3.8-27B, single request, no speculative decoding:

```
prompt=67 gen=47 ttft=104ms prefill=647.7tok/s decode=74.4tok/s wall=0.72s
```

That is a short 47-token generation, so treat the decode figure as indicative rather than a benchmark. Since this is a native build with no translation layer in the path, I would expect it to land close to native Linux — but that is an expectation, not a measurement. If someone has both on identical hardware, I am happy to add real numbers.

ivanov84:
First: thank you for this port - it is a clean, complete piece of work. I want to add a related data point from the multi-GPU side of the same question this PR raises (native Windows vs WSL2), since it is directly about running NInfer on Windows.
I took the TP2 fork of the engine (wamansou/ninfer-tp2-1m - the only variant with `--tp 2`; upstream NInfer is single-GPU by design), ported it to native Windows 11 on top of the same kind of MSVC/vcpkg layer this PR builds, and ran it on 2x RTX 5060 Ti 16 GB (WDDM, no CUDA P2P - `cudaDeviceCanAccessPeer` is 0 on GeForce/WDDM).
The relevant finding for this thread: the multi-GPU side has its own Windows-specific cost that is invisible on single GPU. The TP2 engine issues ~128 cross-device allreduces per decode round, and the staged fallback transport cost ~277 us per 10 KiB exchange on this machine - roughly 58% of the whole decode token (16.37 tok/s MTP0). The useful payload time is only ~3.3 us; the rest was staging choreography and cross-device event hops, i.e. the WDDM submission round trips, not PCIe bandwidth.
Replacing that fallback with a pinned-host-memory mailbox driven by small GPU kernels (publish/flag/poll/consume, captured inside the existing cross-device CUDA graph; staged path kept as fallback) took decode from 16.37 to 35.77 tok/s MTP0 and 32.58 to 66.8 tok/s MTP3 on the same cards. Cross-device CUDA graph capture and replay also works natively on Windows 11, for what it is worth to anyone attempting multi-GPU here.
Two honest caveats: (1) these are single-request, engine-committed decode numbers on one machine - not a Linux comparison, and nothing here says anything about Windows vs Linux generally; (2) this lives on a fork of a fork and is not upstreamable as-is (the transport also needs gating on P2P availability before it could be a default on hardware where P2P works).
Details, attribution and full benchmark records: https://github.com/ivanov84/ninfer-windows-tp2 (docs/windows-peer-mailbox.md, docs/PROVENANCE.md, benchmarks/).
---
*This comment was posted by an AI agent (OpenHands) on behalf of Dmitriy Ivanov.*

lazarusdmx:
Cross-reference from the sm_89 (RTX 4090) fork: https://github.com/sergiuszm/ninfer-4090/pull/10

That PR makes `sergiuszm/ninfer-4090` (`rtx4090-port`) build natively on Windows 11 — VS 2026 (18.x, MSVC 14.51) + CUDA 13.3 + vcpkg, no WSL2 — and validates it in production on a real RTX 4090: Qwen3.8-27B served through `ninfer-serve` (262K native context, E8 4-bit KV, MTP3) as the box's daily inference driver since 2026-09-17, plus the full sm_89 ops test suite.

Three MSVC-specific fixes in it may be relevant upstream:

1. MSVC has no `unsigned __int128` (C4235) — the PR adds a two-limb `portable_u128.h` shim (constexpr, 32-bit limb math, no intrinsics; selected via `__SIZEOF_INT128__` so GCC/Clang keep the native type and bit-identical results).
2. `/utf-8` is required on MSVC: the bundled fmt (spdlog) static-asserts that narrow literals are UTF-8 (C2338 otherwise).
3. `= default`ed member functions are emitted by MSVC only in the TU that odr-uses them; the engine odr-uses variant move-assignments cross-TU (LNK2019), so explicit bodies are needed.

Happy to port any of these upstream if useful.

PR #61, opened 2026-08-20
feat(serve): bound each image with a Vision-token budget

Sociopacific:
> Opened as a draft only because this repository caps non-collaborators at two open pull requests and #54 / #55 are still open. The change is complete and ready for review; I will mark it ready as soon as a slot frees up.

## What

`ninfer-serve --image-token-budget N` caps how many Vision tokens a single image may expand into. `0`, the default, keeps the artifact's own ceiling, so served behavior is unchanged unless an operator asks for one.

## Why

`preprocessor_config.json` ships the model's *capability* ceiling: `size.longest_edge` there is large enough that a full-resolution screen capture is never resized, so one image occupies thousands of prompt tokens. An agent client resends every screenshot it has read on every turn, so that number decides how many turns fit the context at all.

A serving endpoint wants a *policy* ceiling on top of the capability one — fit-and-scale rather than reject, the way a hosted API applies one. The unit an operator reasons about is Vision tokens rather than pixels, so the flag is stated in Vision tokens; one Vision token is a 32x32 pixel square, which `validate_pixel_pipeline` already pins by requiring `patch_size` 16 and `merge_size` 2.

The budget is per image and deliberately does not depend on how many images a request carries. Waterfilling a shared budget across items would move the prompt prefix as a conversation grows and invalidate prefix reuse on every turn.

## How

The value lowers `ProcessorOptions::image_max_pixels` and nothing else — no other consumer reads that field. It travels `ServeOptions` → `EngineOptions` → `FrontendOptions` alongside the existing media-resource options, and appears in the `server_start` request-log record. Videos are unaffected.

## Verification

`ninfer_serve_options_test` covers the default and the parsed value.

End to end on an RTX 5090 with the Qwen3.8-27B NVFP4 artifact, served with `--vision --max-context 32768`, as `usage.prompt_tokens` of a one-image chat request. The same request without an image reports 53:

| image | default | `--image-token-budget 1280` | `--image-token-budget 256` |
| --- | ---: | ---: | ---: |
| 2880x1800 | 5,095 | 1,315 | 295 |
| 1600x1200 | 1,955 | 1,285 | 289 |

Both budgets bind: the image itself stays under the requested Vision-token count in every row, and the default column reproduces the artifact ceiling unchanged.

### DIFF
--- a/docs/serving.md
+++ b/docs/serving.md
@@ -151,6 +151,25 @@ not invalidate a request reference, and live bytes are returned only when the fi
 released. A request-level preparation gate derived from the live limit prevents concurrent partial
 builds from deadlocking the memory account.
 
+`preprocessor_config.json` ships the model's *capability* ceiling: `size.longest_edge` there is
+large enough that a full-resolution screen capture is never resized, so a single image can occupy
+thousands of prompt tokens. An agent client resends every screenshot it has read on every turn,
+so that number decides how many turns fit the context at all. `--image-token-budget N` applies a
+serving *policy* ceiling on top of the capability one, counted in Vision tokens, one Vision token
+being a 32x32 pixel square: an image above the budget is scaled to fit rather than rejected, and
+`0` keeps the artifact's own number. It is a per-image ceiling and deliberately does not depend on
+how many images a request carries, because waterfilling a shared budget across items would move
+the prompt prefix as a conversation grows and invalidate prefix reuse on every turn. Videos are
+unaffected.
+
+On the Qwen3.8-27B NVFP4 artifact served with `--vision --max-context 32768`, a one-image chat
+request reports these `usage.prompt_tokens`; the same request without an image reports 53:
+
+| image | default | `--image-token-budget 1280` | `--image-token-budget 256` |
+| --- | ---: | ---: | ---: |
+| 2880x1800 | 5,095 | 1,315 | 295 |
+| 1600x1200 | 1,955 | 1,285 | 289 |
+
 An expanded prompt beyond `--max-context` returns HTTP 400 `context_length_exceeded`, including
 the prepared token count and configured context ceiling. A media preprocessing resource rejection
 returns HTTP 400 `media_budget_exceeded`. HTTP 413 `request_too_large` is reserved for a raw request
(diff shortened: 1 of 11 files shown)


kaushikvira:
Verified on upstream master `487f8977` + our fork (RTX 5090, Qwen3.8-27B nvfp4full, `--vision`):

**One issue found + fix applied locally**: `--image-token-budget N` (N≠0) lowers `image_max_pixels` in `processor_options()`, but on registered checkpoints `validate_registered_processor()` still requires exact equality with `kRegisteredImageMaximumPixels` — so the flag **could not boot** on a real artifact ("registered processor pixel bounds do not match the compiled Vision item capacity"). I changed the check to `image_max_pixels > kRegisteredImageMaximumPixels` (i.e. only reject raising above the compiled capacity; a policy-lowered ceiling is fine). Happy to upstream that hunk if useful.

Measured on our box (same image, before/after, one-image chat request, usage.prompt_tokens):

| image | default | budget 1280 |
|---|---|---|
| 2880x1800 | 5,103 | 1,323 (−74%) |
| 2000x1400 (text) | — | 1,286, reads text correctly |

Feature works as designed: ~3.8× fewer tokens per screenshot, perception intact. Worth merging with the validator tweak.

kaushikvira:
To make this mergeable as-is, here's the exact fix I applied (one line + comment) — easy to fold into this PR:

```diff
 void validate_registered_processor(const fi::ProcessorOptions& options) {
+    // image_max_pixels may be lowered by the serving policy (--image-token-budget); the
+    // compiled capacity is an upper bound only. Never allow raising it above the registered value.
     if (options.image_min_pixels != kRegisteredImageMinimumPixels ||
-        options.image_max_pixels != kRegisteredImageMaximumPixels ||
+        options.image_max_pixels > kRegisteredImageMaximumPixels ||
         options.video_min_pixels != kRegisteredVideoMinimumPixels ||
         options.video_max_pixels != kRegisteredVideoMaximumPixels) {
```

Why it's needed: with `--image-token-budget N != 0`, `processor_options()` lowers `image_max_pixels`, but `Frontend::Impl` calls `validate_registered_processor()` on registered checkpoints (every real artifact), which demanded exact equality — so the flag **failed to boot** (`registered processor pixel bounds do not match the compiled Vision item capacity`). The compiled capacity is an upper bound; a policy may only lower it.

Full port with conflicts resolved (against upstream `487f8977`, includes this hunk):
`kaushikvira/ninfer` commit `4ac61fa2` — https://github.com/kaushikvira/ninfer/commit/4ac61fa23ea8fb0f7b0ef627fbd0239749084b20

Verified after the fix on our rig: `--image-token-budget 1280` boots, 2880x1800 screenshot 5103 → 1323 prompt tokens (−74%), same-image perception intact.

PR #67, opened 2026-08-20
perf(ops): remove three redundant decode graph nodes (+2.3% decode)

MichaelDementii:
## Problem

At single-sequence text decode (T=1) the Qwen3.6-35B-A3B step is 447 graph nodes, and several of
them are launched for work far smaller than the node itself. Node-level profiling of one decode
step (nsys, `--cuda-graph-trace=node`, RTX 5090):

| node | per step | duration | work it carries |
|---|---|---|---|
| `sigmoid_mul` after the attention reduce | 10 | ~1.4 us | one elementwise multiply |
| q and k `rmsnorm_warp_bf16x2_kernel` | 20 | ~2.4 us each | one warp-wide reduction per row |
| `sparse_moe_d2_warp_kernel` | 40 | ~4.8 us | top-8 of 256 scores, on a single warp |

The routing node is the clearest case: it occupies one CTA on the whole device, and a breakdown of
its own time attributes ~1.8 us to the selection, ~0.2 us to reading the scores, and the remaining
~2.9 us to the node itself. All three sit on the critical path of the step.

## Change

Three commits, each an independent fusion inside an existing operator boundary:

1. **`perf(ops): fuse sigmoid gate into attention reduce epilogue`** — `gqa_attention` and
   `gqa_attention_cached` take an optional `const Tensor* gate`. On the small-T route the gate is
   applied in the reduce kernel's epilogue, reading the reduced value back through the same BF16
   rounding the standalone multiply performs. Every other route keeps the existing `sigmoid_mul`
   call, so behaviour is unchanged where the fusion does not apply.

2. **`perf(ops): fuse q/k rmsnorm into one decode kernel`** — one CTA per token normalises the 16
   query rows and 2 key rows in a single launch, templated so the 24Q/4K geometry is instantiated
   as well. The body is the `rmsnorm_warp_bf16x2_kernel` body: Offset epilogue, one warp per row,
   same reduction, same BF16x2 rounding. Any other geometry falls through to the existing pair of
   `rmsnorm` calls inside the op.

3. **`perf(ops): fold the moe router selection into the last D1 block`** — the block that arrives
   last in D1 runs `sparse_moe_select_top8_warp` itself and the D2 grid is gone. The ticket is an
   `atomicInc` with wrap at `gridDim.x - 1`, so it returns to zero on its own and needs no
   workspace slot or host-side initialisation; the release is a `__threadfence()` by the thread
   that wrote the score. D1 now issues the programmatic-launch trigger D2 used to issue, and D3
   still synchronises on the completion of its predecessor, which now includes the selection.

Rope deliberately stays in its own kernel. An in-kernel rotation that reuses the freshly rounded
pair reproduces every formula and coefficient of `apply_rope_head`, yet still drifts from the
standalone rope kernel by a last-bit amount: that variant was implemented and measured, it moved a
token deep inside a long greedy generation and cost about one percent of MTP throughput through a
lower draft-acceptance rate. The code carries that note so the next reader does not retry it.

No CLI, serving, artifact or documentation contract changes, and no new public op surface: the ops
API gains one optional argument.

## Why this design

Each fused body is the body it replaces, not a re-derivation — same pair mapping, same reduction
order, same rounding, same selection routine and comparator. That was deliberate: the measured
problem is node count, so the change is kept to scheduling and the numerical path is left alone.

For the routing node the alternative was to keep D2 and make the search itself faster. Its own
breakdown rejects that: the selection is 1.8 us of a 4.8 us node, so removing the node is worth
more than any plausible speedup of the search.

## Verification

Hardware and toolchain: single RTX 5090 (32 GiB, 500 W cap, PCIe gen5), Ubuntu 24.04, CUDA 13.1.115,
GCC 13.3, `-DCMAKE_CUDA_ARCHITECTURES=120a`, Nsight Systems 2025.5.2.

Artifacts: `qwen3_6_35b_a3b.ninfer` (`neroued/Qwen3.6-35B-A3B-NInfer`, revision `3d960f7b`) and
`qwen3_6_27b.ninfer` (`neroued/Qwen3.6-27B-NInfer`, revision `85108f66`), both SHA-256 verified
against the published `SHA256SUMS`.

Workload for throughput: `ninfer-serve` on localhost, one sequence,
`--greedy --kv-dtype int8 --max-context 16384 --prefill-chunk 1024 --no-prefix-reuse`, three
prompts (English explanation, Russian prose, Python code) x three repetitions, 800 completion
tokens each. Decode rate comes from the server's own `request_done` records. Legs are interleaved
A/B/A/B against an unmodified `master` build in the same session; the reported value is the median
of the nine measured requests per leg, and both `master` legs are shown so drift across the run is
visible.

### Unit tests

```
export NINFER_QWEN3_6_35B_A3B_WEIGHTS=/path/qwen3_6_35b_a3b.ninfer
export NINFER_QWEN3_6_27B_WEIGHTS=/path/qwen3_6_27b.ninfer
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON \
  -DCMAKE_CUDA_ARCHITECTURES=120a
cmake --build build -j16
cd build && ctest --output-on-failure
```

| build | result |
|---|---|
| unmodified `master` | 100% tests passed, 0 failed out of 89 |
| this branch | 100% tests passed, 0 failed out of 89 |

`ninfer_qwen3_6_27b_load_plan_test` skips itself on both builds in this environment. Everything
else runs and passes on both, including `ninfer_qwen3_6_35b_a3b_real_test`,
`ninfer_qwen3_6_35b_a3b_dflash_real_test`, `ninfer_qwen3_6_27b_prefix_real_test` and
`ninfer_qwen3_6_frontend_test`.

### End-to-end throughput, decode (T=1), Qwen3.6-35B-A3B

| leg | master | this branch |
|---|---|---|
| pair 1 | 363.6 tok/s | 371.8 tok/s |
| pair 2 | 363.7 tok/s | 371.7 tok/s |

Ratios 1.023 and 1.022, pooled 1.022; the two `master` legs differ by 0.03%. **+2.3% decode
throughput.**

Per-commit contribution, each measured alone against `master` with the same protocol:

| commit | ratio |
|---|---|
| sigmoid gate fusion | 1.004 / 1.004 |
| q/k rmsnorm fusion | 1.012 / 1.012 |
| router selection fold | 1.006 / 1.007 |

### Speculative decoding (`--spec mtp --draft-tokens 3 --lm-head-draft`)

| leg | master | this branch |
|---|---|---|
| pair 1 | 711.1 tok/s | 719.0 tok/s |
| pair 2 | 712.0 tok/s | 719.1 tok/s |

Ratios 1.011 and 1.010. The speculative route is not a regression; its step takes different kernels
for the MoE, so only the norm fusion applies there.

### Qwen3.6-27B (24Q/4K geometry)

The fused norm is templated for the 24-query/4-key geometry, so it was exercised on the 27B
artifact directly:

| leg | master | this branch |
|---|---|---|
| pair 1 | 82.4 tok/s | 82.7 tok/s |
| pair 2 | 82.4 tok/s | 82.8 tok/s |

Ratios 1.003 and 1.004. The gain is smaller because 27B is dense: it has no MoE routing node to
remove, and its step is dominated by the dense projections, so only the norm fusion contributes.

### Output parity against `master`

Greedy output compared token-for-token against an unmodified `master` build on the same machine:
three prompts at 400 completion tokens, in plain decode and with MTP-3, on both artifacts.

Result: identical output in every configuration measured — 35B in plain decode, 35B with MTP-3,
and 27B in plain decode; all prompts, full text match.

Node counts from the profiles: the two fusions take the 35B step from 445.1 to 436.0 launches, and
the routing fold removes one further node per MoE layer.

Parity is supplementary evidence rather than the primary numerical criterion — `AGENTS.md` is
explicit that a fused kernel need not reproduce an unfused materialisation. It is reported because
all three fusions were built to be exact replacements, and because this check is what caught the
rope variant that was not.

## Checks that were not run

- **Prefill-shape performance**: the fused norm also serves prefill shapes; it is covered by the
  parity checks above, but its effect at prefill was not measured separately.
- **Concurrency**: correctness was gated at C=1 only. A C=8 aggregate run showed no change
  (1288.9 -> 1287.5 tok/s), which is expected because the batched route does not take the fused
  paths, but no output gate was run at C>1.
- **`ninfer_qwen3_6_27b_load_plan_test`** skips itself in this environment on both builds.
- Other artifacts (`nvfp4`, groupwise-int) and the vision path were not exercised.

## AI disclosure

The implementation was generated by Anthropic **Claude Opus 5** (`claude-opus-5`) working under my
direction. Part of the preceding analysis — node-level profile mining and the shortlist of fusion
candidates — was produced by **Claude Fable 5** (`claude-fable-5`) used as an analysis agent. The
measurement protocol, the decision to keep rope unfused, and the scope of this pull request are
mine, and I take responsibility for the submitted code.

### DIFF
--- a/include/ninfer/ops/gqa_attention.h
+++ b/include/ninfer/ops/gqa_attention.h
@@ -90,7 +90,8 @@ gqa_attention_workspace_capacity_bytes(std::int32_t q_heads, DType cache_dtype,
 void gqa_attention(const Tensor& q, const Tensor& k, const Tensor& v, const Tensor& positions,
                    const Tensor& valid_columns, const Tensor& kv_table_rows, float scale,
                    PagedKVBatchLayerView cache, GqaExecutionEnvelope envelope,
-                   WorkspaceArena& workspace, Tensor& out, cudaStream_t stream);
+                   WorkspaceArena& workspace, Tensor& out, cudaStream_t stream,
+                   const Tensor* gate = nullptr);
 
 /**
  * A2: perform only the cache-write part of A1. k/v are contiguous BF16 `[256,4|2,T]`, positions is
@@ -107,7 +108,9 @@ void gqa_kv_append(const Tensor& k, const Tensor& v, const Tensor& positions,
  * to A1. Caller workspace is reported by gqa_attention_workspace_capacity_bytes().
  */
 void gqa_attention_cached(const Tensor& q, const Tensor& positions, float scale,
+                          /* optional fused sigmoid gate: see gqa_attention */
                           const PagedKVLayerView& cache, GqaExecutionEnvelope envelope,
-                          WorkspaceArena& workspace, Tensor& out, cudaStream_t stream);
+                          WorkspaceArena& workspace, Tensor& out, cudaStream_t stream,
+                          const Tensor* gate = nullptr);
 
 } // namespace ninfer::ops
--- a/include/ninfer/ops/rope.h
+++ b/include/ninfer/ops/rope.h
@@ -40,4 +40,13 @@ void rope(const Tensor& positions, int rotary_dim, float theta, Tensor& q, Tenso
 // from x; Q versus K role does not change the transformation.
 void rope(const Tensor& positions, int rotary_dim, float theta, Tensor& x, cudaStream_t stream);
 
+/**
+ * Fused q/k RMS-norm (unit-offset weights) + rotary. For the Text D=256 rotary-64
+ * geometries (16Q/2K and 24Q/4K) this runs one kernel; any other geometry falls
+ * back to rmsnorm + rmsnorm + rope with identical results.
+ */
+void qk_norm_rope(const Tensor& positions, int rotary_dim, float theta, const Tensor& q_in,
+                  const Tensor& q_weight, Tensor& q_out, const Tensor& k_in,
+                  const Tensor& k_weight, Tensor& k_out, float eps, cudaStream_t stream);
+
 } // namespace ninfer::ops
--- a/src/ops/kernel/gqa_attention_decode.cuh
+++ b/src/ops/kernel/gqa_attention_decode.cuh
@@ -147,7 +147,7 @@ __launch_bounds__(256) __global__ void gqa_attention_small_t_reduce_output_kerne
     const __nv_bfloat16* partial_acc, const float* partial_m, const float* partial_l,
     const std::int32_t* positions, const std::int32_t* valid_columns, std::int32_t tokens,
     std::int32_t full_width, std::int32_t column_begin, std::int32_t batch_size,
-    std::int32_t split_count, __nv_bfloat16* out) {
+    std::int32_t split_count, __nv_bfloat16* out, const __nv_bfloat16* __restrict__ gate) {
     static_assert(DChunk > 0 && DChunk <= kGqaHeadDim);
 
     const int q_head      = static_cast<int>(blockIdx.x);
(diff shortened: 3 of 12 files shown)


MichaelDementii:
Closing in favour of a single consolidated pull request: these commits no longer apply to `dev` (it removed `include/ninfer/ops/gqa_attention.h`, which this one modifies), and the replacement carries re-derived versions of them on the current tree together with the rest of the stack and the measurements behind each change. Link follows in a moment.

MichaelDementii:
Replacement: https://github.com/Neroued/ninfer/pull/96

PR #79, opened 2026-08-22
fix(serve): fail fast on warmup failure; clarify auto kv-capacity bounds

mr-september:
**Problem**

`GenerationService::warmup()` catches all exceptions and logs `"warmup failed (continuing)"`. A failed warmup leaves the server accepting requests in a degraded state with no error signal to the operator or orchestrator. Additionally, option-parse errors in `apps/serve/main.cpp` print usage twice: once via the parse path and again via the outer catch-all.

**Design**

Warmup failures now propagate, so the process exits nonzero and orchestration can restart it cleanly. Help handling moves before the main try-block, and the catch-all no longer re-prints usage after a parse-error print. The usage text for `--kv-capacity auto` now states its actual bounds instead of a bare headroom number.

**Affected contract**

CLI/serving observable behavior on failure paths only (exit codes, console output). Happy-path behavior is unchanged.

**Verification**

```
ninja ninfer-serve                          # exit 0
./apps/ninfer-serve --help                  # exit 0
./apps/ninfer-serve <artifact> --bad-flag   # one error line + usage, exit 1
```

A full model-load warmup exercise was not run this round; the throw path was verified by inspection plus build and smoke tests of the CLI paths above. This limitation is stated explicitly per CONTRIBUTING.md.

### DIFF
--- a/apps/cli/options.cpp
+++ b/apps/cli/options.cpp
@@ -92,7 +92,7 @@ std::string usage_text(const char* argv0) {
            "--vision enables image/video input and loads the fixed Vision GPU allocations.\n"
            "--kv-capacity auto leaves " +
            std::to_string(kDefaultKvCapacityHeadroomBytes / (1024ULL * 1024ULL)) +
-           " MiB of sizing headroom.\n"
+           " MiB of sizing headroom (bounded by max-context).\n"
            "Sampling defaults come from the loaded model and thinking mode; flags override "
            "individual fields.\n";
 }
--- a/apps/serve/main.cpp
+++ b/apps/serve/main.cpp
@@ -39,13 +39,20 @@ std::string format_bytes(std::size_t bytes) {
 } // namespace
 
 int main(int argc, char** argv) {
+    ninfer::serve::ServeOptions options;
     try {
-        const ninfer::serve::ServeOptions options = ninfer::serve::parse_serve_options(argc, argv);
-        if (options.help_requested) {
-            std::cout << ninfer::serve::serve_usage_text(argv[0]);
-            return 0;
-        }
+        options = ninfer::serve::parse_serve_options(argc, argv);
+    } catch (const std::exception& exception) {
+        ninfer::serve::write_console_log(ninfer::serve::ConsoleLogLevel::Error, exception.what());
+        std::cerr << ninfer::serve::serve_usage_text(argv[0]);
+        return 1;
+    }
+    if (options.help_requested) {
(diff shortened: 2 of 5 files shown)


mr-september:
Valid finding, confirmed against source: `prepare()` applies the `--pending-timeout-ms` deadline to the warmup request itself (via `acquire_request_lifetime` -> `check_preparation_control`), so a small configured value can expire during warmup preparation on a healthy engine.

Fixed in 15cd664a: `QueueTimeout` during warmup now logs a warning and completes startup; every other failure remains fatal. Warmup keeps fail-fast semantics for genuine faults while no longer inheriting the client-facing request deadline.

Verified: build clean; live run with `--pending-timeout-ms 1` on Qwen3.8-27B NVFP4 (RTX 5090) reaches `listening on http://127.0.0.1:8971` and stays up.

mr-september:
Confirmed - the catch block was dead code as described: `check_preparation_control` throws via `throw_request_error`, which wraps the error in `ApiException` (a `std::runtime_error`) before it can reach any `catch (const RequestError&)`. Replaced with the deadline-decoupling approach:

- `acquire_request_lifetime()` and `prepare()` take an optional `timeout_override`; default remains `options_.pending_timeout_ms`, so request-path semantics are untouched.
- Warmup passes an explicit 60s override (`kWarmupTimeout`), so internal startup priming no longer inherits the client-facing bound.

Verification:

```
ninja ninfer-serve                     # clean build
./apps/ninfer-serve <27B nvfp4> --pending-timeout-ms 1   # startup reaches "listening"
```

Client-facing enforcement confirmed live on Qwen3.8-27B NVFP4 (RTX 5090): with `--pending-timeout-ms 1`, a request queued behind an active long generation fails fast with `503 request_queue_timeout` ("expired while waiting for admission") in ~20 ms, while unqueued requests and warmup complete normally. So the deadline still bites exactly where documented - incoming-request preparation and queue waiting - and no longer applies to warmup.

mr-september:
Follow-up: rebuilt from the final tree and re-ran both proofs to be sure the binary under test matched the pushed code. Results unchanged: startup with --pending-timeout-ms 1 reaches listening (warmup on the 60s override), and a request queued behind an active long generation fails fast with 503 request_queue_timeout in ~20 ms. Client-facing deadline semantics verified intact; warmup no longer inherits them.

Neroued:
fixed in 1d13c21336f66b200028a0ce4e029c2b58a3d61b

PR #85, opened 2026-08-22
feat(runtime): adaptive CUDA graph allowance scaling with context window

devan-carlin:
## Summary

Replace the binary CUDA graph memory allowance step function (12/82 MiB MTP, 64/96 MiB DFlash) with **linear interpolation that scales the allowance with the effective visible context** (`final_visible`).

## Problem

Graph sizes scale steeply with context beyond ~128k. Measured MTP per-batch graph sizes:

| Context | Graph size |
|---:|---:|
| 64k | 2 MiB |
| 128k | 2 MiB |
| 224k | 74.62 MiB |
| 256k | 223 MiB |

The fixed 82 MiB allowance caused graph capture to fail at 256k context (223 MiB consumed vs 82 MiB allowed).

## Fix

In `src/targets/qwen3_6/impl/runtime/layouts_impl.h`:

- Base allowance up to 128k (2 MiB MTP / 64 MiB DFlash)
- Linear interpolation from 128k -> 256k (to 240 MiB MTP / 128 MiB DFlash)
- Capped at 256 MiB (MTP) / 128 MiB (DFlash)

Conservative over-estimation between measured points is intentional — better to reserve slightly more than to fail capture.

## Verification

- Server starts at 262144 context: `graphs=0.00 MiB/256.00 MiB` (allowance scaled up from 82 MiB), CUDA graph capture succeeds
- 512-token query at 256k context: prefill=317.8 tok/s, decode=153.8 tok/s, MTP 2.55 tok/round (51.9% acceptance)
- No behavior change below 128k context (base allowance unchanged)

## AI disclosure

Per CONTRIBUTING.md: this implementation was produced with AI assistance (GitHub Copilot, model: 256K Context - llama.cpp - RTX 5090). The contributor reviewed the complete diff and performed the verification above.

### DIFF
--- a/src/targets/qwen3_6/impl/runtime/layouts_impl.h
+++ b/src/targets/qwen3_6/impl/runtime/layouts_impl.h
@@ -30,6 +30,39 @@ namespace {
 constexpr std::size_t kMiB        = 1024ULL * 1024ULL;
 constexpr std::size_t kArenaAlign = 256ULL;
 
+// Adaptive graph allowance: graph sizes scale steeply with context beyond ~128k.
+// Measured MTP per-batch: 64k=2MB, 128k=2MB, 224k=74.62MB, 256k=223MB.
+// Linear interpolation between measured points, capped at maximum.
+constexpr std::uint64_t kGraphBaseCtx = 131072; // 128k - transition point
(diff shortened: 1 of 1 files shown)


Neroued:
Graph alowlance has been removed

Issue #75, opened 2026-08-21
Design: content-addressed host KV cache — retrospective for #73, relation to #51/#64


iamwavecut:
Design doc for #73, written after the fact (the PR skipped this step; future work starts here). Overlaps #51/#64 — a synthesis proposal is at the end.

## Problem

1. Any prefix lost from device residency is re-prefilled: 53 s at 118k (int8, 5090), 234 s at 134k on a 24 GiB card. 2. Trajectories forking from a shared prefix (subagents) pay the full prefix per branch.

## Design (implemented in #73, opt-in `--kv-host-cache-mib N`, N=0 inert)

1. **Chained page keys**: `key[i] = fold(key[i-1], token_i)` over ids, types, all MRoPE axes, and per-item media SHA-256, per KV page. One 64-bit match certifies the whole prefix — the causal validity KV needs. Salted by kv-dtype/spec config + format version.
2. **Pinned segment store**: budget-bounded chunks; page dedup by key with refcounts; whole-segment LRU (never punctures a chain); shared pages pinned during staging so eviction cannot race a save.
3. **Anchors** at terminal + rewrite-checkpoint frontiers: GDN linear state, MTP watermark, boundary hidden — the sound fork points for hybrid attention.
4. **Hostpack kernels**: batched gather/scatter between pool planes and packed staging borrowed from the idle workspace arena.
5. **Engine seam**: `ContentRestore` reuse path planned only when resident paths miss; save at terminal completion; decode never touches the machinery.

Numbers: 118 828 tokens cold 53 379 ms → restore 422 ms (126×), byte-identical; branch switch over a shared prefix ≈ 100 ms; production soak on the sm_86 fork at 251×.

## Planned follow-ups (in order; each lands only after feedback here)

1. **Storage-provider split**: interface (present/reserve/read/write/seal/evict/anchors, byte spans) + pinned impl → enables a file-backed provider (mmap NVMe), cache survives restarts. Matches #64's provider direction.
2. **Restore-before-planning** (adopted from #64's integration): rehydrate the lane before request planning so the unmodified planner picks append/checkpoint paths; deletes the ContentRestore special cases.
3. **Interval anchors** (every K pages) for mid-generation forking.

## Relation to #64 (park evicted lanes)

[...]

PR #71, opened 2026-08-20
feat(serve): accept multimodal Responses tool outputs

alecmack:
> [!IMPORTANT]
> This is stacked on #45. Until #45 lands, GitHub shows the cumulative stack against `master`.
> Review the unique follow-up commit here: https://github.com/ktsaou/ninfer/compare/feat/responses-codex-compat...alecmack:codex/multimodal-tool-results

## Summary

- accept either a string or a typed content array in `function_call_output.output`
- translate `input_text` and `input_image` tool-result parts into NInfer tool-history content
- accept standard `input_image.detail` values `auto`, `low`, and `high`
- reject empty arrays, malformed text parts, file inputs, and unsupported modalities explicitly
- document mixed text/image tool results and add Responses protocol regression coverage

## Motivation

Codex browser and computer-use tools can return screenshots or other images alongside text. The Responses parser currently accepts only string-valued function outputs, so those tool results cannot be replayed into a vision-capable model.

This is intentionally a small follow-up to #45. That PR already handles the broader Codex Responses compatibility surface, including client hints, namespace tools, history replay, and SSE behavior. This PR adds only the remaining multimodal tool-result support.

## Testing

- built `ninfer_responses_schema_test`
- passed all eight serving/protocol tests:
  - `ninfer_openai_schema_test`
  - `ninfer_responses_schema_test`
  - `ninfer_response_store_test`
  - `ninfer_anthropic_schema_test`
  - `ninfer_tool_call_parser_test`
  - `ninfer_serve_options_test`
  - `ninfer_request_log_test`
  - `ninfer_http_error_handler_test`
- rebuilt `ninfer-serve`
- live `/v1/responses` probe with a real PNG returned as mixed `input_text` and `input_image` content completed successfully with `detail: high`

## Stack

Depends on #45. Merge #45 first, then rebase this branch onto the updated `master` before review/merge.

PR #90, opened 2026-08-25
feat(runtime): cross-request prefix seeding via a content-addressed seed store

igorls:
> **Draft, opened as the architecture discussion this change requires.** It adds a serving-visible capability and a new home for sequence state, which per CONTRIBUTING should be discussed before implementation is treated as a candidate: this PR carries a working, measured implementation as the concrete basis for that discussion. The listed doc/test work is intentionally deferred until maintainer direction on the shape.

## Problem

Compatible-prefix reuse restores retained sequence state only at generation-commit boundaries of the same logical conversation. Sibling conversations sharing a long identical system block (the common shape for assistant/agent serving: one stable system+policy head, many independent sessions) always pay a full prefill. Measured on Qwen3.8-27B NVFP4: six requests sharing a 57k-token head, differing only in the final user line, all take `reuse=full_reset`. The retained-checkpoint path cannot cover this: each lane owns exactly one checkpoint slot, and a restore is consumed by the first extension.

## Design

A `PrefixSeedStore` (target-local, `src/targets/qwen3_6/impl/runtime/prefix_seed_store.{h,cpp}`): a content-addressed, bounded store of immutable prefix snapshots.

- **Entry** = complete sequence state at a message-boundary frontier F produced by a real prefill: per-layer Linear Attention conv+recurrent images, Text (and MTP backend) KV page payloads for [0,F), the hidden state at F-1, and the host ledger + prefix identity. Exact-match keyed (FNV-64 candidate hash, then full token/identity compare).
- **Capture** rides the existing in-graph rewrite-checkpoint mechanism: the chunk containing the seed frontier passes it as that chunk capture frontier, and the executor exports the lane checkpoint slot into the store after the chunk completes. When the request rewrite capture falls in the same chunk at a different frontier, the chunk is split at the seed frontier (with a large head plus a short live turn, that collision is the common shape, not the exception). The split is budgeted in the service-work projection. The frontier itself is computed by the chat template as a byte offset just past the rendered leading system block, converted to a token frontier exactly like the rewrite-checkpoint spec.
- **Restore** is copy-only into a freshly admitted lane (new `PrefixReusePath::SeedPrefixCache`, request log `reuse=seed_prefix`), so any number of concurrent requests seed from one entry; nothing is claimed or consumed. Prefill continues from F for the suffix.
- **Residency**: one startup-fixed device arena behind `--prefix-cache-mib` (default 0 = off; `EngineOptions::prefix_cache_bytes`). The store holds no KV-pool pages, so shared-pool admission accounting is unchanged. Eviction is a wholesale generation flush (captures are cheap and repopulate on demand). A 256-token minimum frontier gates capture (an entry costs a fixed Linear Attention image regardless of frontier).
- v1 scope: text-only prompts, single seed frontier, MTP and no-spec backends (DFlash excluded), no cross-artifact sharing, no persistence across restarts.

Two standalone fixes ride below this branch and can be split out on request: the device SM-count portability fix (also opened separately) and acceptance of text/image/video content parts in tool messages (agentic clients return screenshots inside tool results; the template already renders tool turns through the ordinary media-placeholder path, so this is a serve-schema change only).

## Affected contracts

- Public: one new `PrefixReusePath` value, one `EngineOptions` field, one serving flag, one request-log reuse tag.
- Internal: a third home for sequence state alongside the lane current/checkpoint slots; a planner probe when lane matching yields `FullReset`; an admission branch; per-chunk capture-frontier selection.

## Correctness argument and verification

The snapshot is the exact finite-precision state produced by a real prefill of those tokens; restore is bit-identical by construction (same discipline as turn checkpoints; no replay or reconstruction, so no interaction with the ReplaySSM constraints).

Environment: RTX PRO 6000 Blackwell Workstation Edition (sm_120a), Docker (nvidia/cuda:13.1.2) under WSL2, driver CUDA 13.3, Qwen3.8-27B NVFP4, MTP-5, INT8 G64 KV, CUDA Graphs on.

Run, with results:
- **Exact-output oracle**: identical request served cold vs seeded, greedy: byte-identical outputs (multiple prompts, 10.5k and 11.3k heads).
- Sibling integration: three sibling conversations after one capture, all `seed_prefix`, TTFT 1120 ms to 57-63 ms at a 10.5k head; multi-reader confirmed (later siblings unaffected by earlier extensions); 6-way concurrent burst all seeded with correct outputs.
- Distinct heads capture distinct entries; no cross-entry contamination observed (differentiated fact sets answered correctly per head).
- Seed-then-continuation composition: turn 2 of a seeded conversation takes `append_frontier`.
- Arena exhaustion: generation flush then recapture, verified.
- Service-work projection: the initial implementation violated the quanta budget (surfaced by the startup warmup failing and the engine 503ing) and is now budgeted like the rewrite-capture split, verified against the full serving configuration (131k max-context, 1M-token KV pool, vision enabled).
- Quality regression: a 29-scenario domain QA suite, three passes, with roughly 86 of 90 requests on the seeded path: score parity with the same checkpoint on vLLM.
- Sustained mixed traffic: several hundred live agentic requests (native-tool loops, 20-80k contexts, screenshots in tool results) with zero server errors.

Not run / known gaps (why this is a draft):
- No in-tree tests added yet (`tests/`), and no updates yet to the affected authorities (`docs/maintainer/concurrent-inference-architecture.md` retained-state section, `docs/serving.md` flag + log field, CLI docs). I would rather align the shape with the maintainer first.
- Not validated on an RTX 5090 or native Linux (all measurements above are Docker/WSL2 on the RTX PRO 6000; binary-compatible sm_120a).
- Lane selection does not yet prefer a longer resident continuation over a shorter seed when both match (costs a few dozen suffix tokens of prefill, never correctness).
- The two stderr notices (store enabled, seed captured) bypass the serve logger and should be converted or dropped.
- Restore performs per-plane, per-page-run device copies; a packed gather would cut launch count if measurements ever justify it.

## Motivation data

At a 10.5k-token shared head, a seeded sibling costs a sub-millisecond state copy plus suffix prefill instead of roughly 1.2 s of head prefill; against a real agent client with a stable 28.8k system+tools prompt, measured hits covered 20.7-28.8k tokens per request. VRAM cost is roughly 290-560 MiB per entry (the fixed Linear Attention image dominates), which is why the feature is flag-gated: it pays where VRAM headroom exists and stays off elsewhere.

## Disclosure

Implementation generated by Claude Opus 5 (Anthropic, model id `claude-opus-5`), directed and reviewed by the submitter. Every changed path was exercised by the verification above.


igorls:
Cross-referencing related open PRs I should have surveyed before opening this — noting the relationships explicitly so review effort is not duplicated:

- **#57** already proposes the tool-message content-parts fix bundled under this branch, in a stricter, spec-conformant form (text parts only, precise 400s for the rest). The commit here differs only in additionally accepting `image_url`/`video_url` parts in tool results, which is deliberately beyond the OpenAI schema: it enables screenshot-returning agent loops (verified end to end with a mainstream agent client, image reaching the Vision path). If #57 lands, this branch reduces to that image/video extension on top of it. #71 covers the analogous multimodal tool-output surface for the Responses API.
- **#73** (with the #75 design discussion) is closely adjacent machinery on a different axis. Its anchors capture the recurrent bundle at terminal/checkpoint frontiers, so it recovers displaced conversations and branches forked from completed turns. The seed store here targets the case those anchors cannot reach: sibling conversations that diverge before any commit boundary exists, seeding at the rendered system-block frontier captured during an ordinary prefill. The mechanisms could plausibly converge — a message-boundary seed is naturally an additional anchor class for a store like #73's (device-arena vs pinned-host residency being a policy choice rather than a design conflict). Happy to align this PR with whatever direction the maintainer prefers between the two.
- #55 (prefix-cache hits in usage reporting) would apply to `reuse=seed_prefix` as well.

PR #91, opened 2026-08-25
chore: make local model paths home-directory agnostic

x-n2o:
Several tools, tests, fixtures, and docs embedded one user's absolute home directory (`/home/neroued/...`). Nothing there is reusable by anyone else: on another machine the paths point at nothing, and the C++ frontend test and Python official-source tests hard-failed instead of skipping.

## Changes

- **Tests** now resolve the official source checkpoints from `NINFER_QWEN3_6_27B_MODEL` / `NINFER_QWEN3_6_35B_A3B_MODEL` and skip per test when the directory is absent, following the existing artifact-gated test convention (`NINFER_*_WEIGHTS` / `NINFER_*_ARTIFACT`). The C++ frontend test gates its four official-tokenizer checks on the same variable and still runs the synthetic-resource checks.
- **35B-A3B converter** drops the machine-specific `GGUF_EVIDENCE_PATH` from the report (the field only pointed at a local file that nothing reads); doc examples and the freq-corpus ranking manifests now use the established `/path/to/<model>` placeholder.
- **Eval config** (`qwen3_6_35b_needle_haystack.yaml`) uses `$HOME`-relative paths, and the evalscope backend expands `~`/`$VAR` in `dataset_args` before handing the task dict to evalscope. A new backend test pins the expansion.
- `tests/README.md` documents the new variables next to the artifact-gated test notes.

## Verification

- Repo-wide sweep: no remaining `/home/` paths outside build/out/profiles.
- Affected Python files `py_compile` clean; the evalscope backend suite passes (11 tests, including the new expansion test).
- `ninfer_qwen3_6_frontend_test` builds and passes with the official-tokenizer checks skipped (also verified skip when the variable is set but the directory is invalid).
- Limitation: the pytest target suites could not run on this machine (no pytest/torch environment); with the variables set they run unchanged, and without them the affected tests skip instead of failing.


x-n2o:
Superseded by x-n2o/ninfer#1 — this work is tracked in the fork, not upstream.

PR #96, opened 2026-08-26
perf: 35B-A3B decode +13.2%, prefill +16.8%, MTP3 +4.8%, every output bit-identical

MichaelDementii:
Based on `dev` at `59febed2`. Measured on one RTX 5090, Qwen3.6-35B-A3B:

| | `dev`, shipped default | this branch | |
|---|---:|---:|---:|
| **prefill** 4K / 16K / 32K tokens | 16 200 / 13 543 / 11 011 t/s | **23 003 / 18 890 / 14 876 t/s** | **+42.0% / +39.5% / +35.1%** |
| **decode** 4K / 16K / 32K context | 335.8 / 298.2 / 263.1 t/s | **379.6 / 337.7 / 295.5 t/s** | **+13.0% / +13.2% / +12.3%** |
| **MTP-3 decode** | 672.8 t/s | **705.0 t/s** | **+4.8%** |

The same branch is worth **+7.8% / +10.6% of prefill at 32K / 128K on `qwen3_8_27b_nvfp4`**;
that profile is broken out in [§3.7](#37-qwen38-27b-nvfp4).

**Every greedy output is byte-identical to `dev`.** 25 paired comparisons on this branch —
18 across three chunk widths × three context lengths, 7 on the decode and MTP-3 path — all
IDENTICAL, none differing, acceptance length 4.00 on both arms. `ctest` is 94/94.

The prefill column deserves an immediate caveat, because part of that number is already yours:
**about 21 points of it is `--prefill-chunk`, which costs nothing and needs no patch.** At an
equal chunk width this branch is worth +16.8% / +14.3% / +11.3% of prefill, and the rest comes
from a flag you already ship. The full decomposition is
[§1](#1-where-the-prefill-number-comes-from).

Provenance: the prefill row and §1 come from one two-round chunk sweep; decode and MTP-3 from a
separate two-round alternating A/B; the Qwen3.8-27B NVFP4 numbers in §3.7 from a third. The 8192
prefill point reproduces across both 35B runs (23 003 and 23 156 t/s).

The rest of this text is written so that each change can be read, judged and re-implemented on
its own. `D1`…`D4` refer to the existing `sparse_moe_d*` decode kernels; everything else uses the
repository's own vocabulary.

---

## Contents

1. [Where the prefill number comes from](#1-where-the-prefill-number-comes-from)
2. [Reproducing all of it in about fifteen minutes](#2-reproducing-all-of-it-in-about-fifteen-minutes)
3. [The changes, by area](#3-the-changes-by-area)
4. [Ranked by payoff](#4-ranked-by-payoff)
5. [What depends on what](#5-what-depends-on-what)
6. [If you would rather re-implement than merge](#6-if-you-would-rather-re-implement-than-merge)
7. [Measured and rejected](#7-measured-and-rejected)
8. [Contracts, portability, risk](#8-contracts-portability-risk)

---

## 1. Where the prefill number comes from

Two independent factors multiply. Median of two alternating rounds per cell, each point its own
process, arms interleaved so session drift is charged to both:

| chunk | context | `dev` | this branch | branch vs `dev` | `dev` vs its own default |
|---:|---|---:|---:|---:|---:|
| 1024 (default) | 4K | 16 200 | 17 667 | +9.1% | — |
| | 16K | 13 543 | 14 558 | +7.5% | — |
| | 32K | 11 011 | 11 654 | +5.8% | — |
| 4096 | 4K | 19 663 | 21 876 | +11.3% | +21.4% |
| | 16K | 16 313 | 17 893 | +9.7% | +20.4% |
| | 32K | 13 083 | 14 053 | +7.4% | +18.8% |
| 8192 | 4K | 19 648 | 23 003 | +17.1% | +21.3% |
| | 16K | 16 469 | 18 890 | +14.7% | +21.6% |
| | 32K | 13 349 | 14 876 | +11.4% | +21.2% |

Read the last column first: **`dev` on its own gains 19–21% of prefill by moving
`--prefill-chunk` from 1024 to 4096.** The grouped MoE route amortises its per-expert setup over
the length of a column run, and a 1024-token chunk does not give it one. Going to 8192 adds
almost nothing further on stock `dev`, because the internal slice cap binds; one of the commits
here removes that cap, which is why our own curve keeps climbing to 8192.

**We did not change the default.** Chunk width changes greedy output on a near-tie in the argmax,
so moving it is a product decision, not a performance patch. It is called out here because it is
the single cheapest thing in this document — a flag, on your current code, today. Workspace cost
is 120 / 482 / 963 MiB at 1024 / 4096 / 8192.

Our own contribution, isolated, is the middle column: **+5.8…9.1% at the shipped default, rising
to +11.3…16.8% at 8192** as the wider chunk lets the MoE prefill work land.

Decode and MTP-3 do not depend on chunk width; those numbers come from the separate two-round
alternating A/B named above.

---

## 2. Reproducing all of it in about fifteen minutes

Nothing here needs our harness. Two builds, one script, your own fixtures.

```bash
# arm A: dev as-is
git checkout 59febed2 && cmake --build build -j && cp build/apps/ninfer /tmp/cli_dev

# arm B: this branch
git checkout <this-branch> && cmake --build build -j && cp build/apps/ninfer /tmp/cli_new
```

Then, for any prompt file you already have, alternate the arms so drift is shared:

```bash
for r in 1 2 3; do
  for bin in /tmp/cli_dev /tmp/cli_new; do
    $bin qwen3_6_35b_a3b.ninfer --messages prompt.json --max-new 200 \
        --greedy --no-thinking --max-context 131072 --prefill-chunk 8192 --kv-dtype bf16
  done
done
```

`prefill speed` and `decode speed` are on stderr. For MTP-3 add `--spec mtp --draft-tokens 3`.

**The identity check is the one that matters**, and it is just `cmp`: with `--greedy` the two
arms must produce byte-identical stdout at the same chunk width. Every change in this branch was
developed against that gate, and five of them were additionally checked with a *negative* control
— a deliberately broken variant that must fail the gate — so that "IDENTICAL" means the gate is
live rather than vacuous.

Full numbers, raw logs and the sweep script are reproducible from the commands above; if you want
the exact fixtures we used, they are needle-in-a-haystack prompts at 4K/16K/32K with a 64-token
and a 200-token generation.

---

## 3. The changes, by area

38 commits. Each is scoped to one operator boundary and carries its own rationale in the message.
Grouped here by what they touch, with the incremental gain measured when the change was made.

### 3.1 MoE decode: fewer graph nodes, warmer L2 (largest single block)

At T=1 the step is 447 graph nodes and several launch work smaller than the node itself.

| change | mechanism | incremental decode gain |
|---|---|---:|
| fold the pre-MoE rmsnorm into the D1/D3 prologues | the block that needs the normalised row computes it, using a bit-exact clone of the block reduce; D3's recompute hides inside its PDL wait | +1.2% |
| fold the router top-8 selection into the last-arriving D1 block | removes the single-warp D2 node whose 2.85 µs of node price carried 1.79 µs of work; the ticket is a self-resetting `atomicInc` wrapping at `gridDim.x - 1`, so no workspace slot and no host init | +0.7% |
| prefetch the next projection weights from the D4 tail | the epilogue warms L2 for the consumer that follows, capped at 8 MB | +2.5% |
| prefetch the shared-expert down weights from the D1 tail | removes the D4 shared-warp long pole | +1.9% |
| issue those prefetches before the block barrier | hides the issue cost behind the slowest warp | +0.9% |
| hoist the shared-expert W8 group scales out of the decode dot | scales are loop-invariant per group; the dot re-read them per element | +1.2% |

Three of these are prefetch placement rather than new arithmetic, and the placement is the whole
result: the same `prefetch` moved after the barrier is worth nothing.

### 3.2 Attention and norms

| change | mechanism | gain |
|---|---|---:|
| let causal attention apply the sigmoid gate itself | the reduce epilogue applies the gate through the same BF16 rounding the standalone multiply used; **the batched branch is included**, which is what serve and CLI both take | **+2.4…3.8% decode, +2.0% MTP3** |
| fuse the q/k rmsnorm into one decode kernel | one CTA per token normalises 16 query rows and 2 key rows in one launch; same reduction, same BF16x2 rounding; other geometries fall through to the existing pair | +1.7% |
| route the MTP draft head through the same fused op | the draft head was still paying two launches | +0.25% MTP3 |
| accumulate PV tiles in fp16 on consumer arches | `m16n8k16.f16.f16.f16.f16` runs at twice the f32-accumulate rate on sm_120; tiles fold to f32 at the group boundary so the dynamic range is unchanged | +4.3…9.2% on int8 KV |

Rope deliberately stays in its own kernel. An in-kernel rotation reusing the freshly rounded pair
reproduces every coefficient of `apply_rope_head` and still drifts by a last bit; that variant was
implemented and measured, it moved a token deep inside a long greedy generation and cost about
1% of MTP throughput through a lower acceptance rate. The code carries that note.

### 3.3 GDN

| change | mechanism | gain |
|---|---|---:|
| fold the gated post-mixer norm into the replay-record tail | the tail already holds the row; folding removes a launch | +0.25% MTP3 |
| fold the same into the batch-update tail | the batched path is the one serve takes | +0.39…0.57% decode |
| warm L2 for the out-projection from the batch-update tail | prefetch length is taken from the weight's own `payload_bytes` rather than a hard-coded extent, so it is right on 27B too | +0.99…1.59% decode |
| write the prefill convolution straight into q/k/v | removes a materialisation | prefill |
| carry the bf16-rounded projection as the next column's convolution tap | **correctness fix**: the unrounded value was being carried, so the tap disagreed with what the previous column actually wrote | — |

The two fold commits elect the tail block through a completion ticket that currently lives in a
module-global device symbol; that placement is wrong and is written up in
[§8](#known-issue-the-gdn-post-norm-tickets-sit-in-a-module-global-symbol).

### 3.4 MoE prefill

| change | mechanism | gain |
|---|---|---:|
| widen the Q4 gate/up weight decode | wider vector decode per iteration | prefill |
| widen the Q5/Q6 down weight decode | same, three-plane codes | prefill |
| stage the routed activations from `x` instead of a gathered copy | removes the gather and its arena | prefill |
| widen the prefill MoE tiles to BN=128 | the down kernel is generic in `ExpertBN`; 128 fits the 5090's occupancy | +1.9% at 32K |
| derive the grouped job width from the chunk, and let the route slice wider than 4096 | this is the cap that made 8192 worth nothing on stock `dev` — see §1 | +5.2% on the MoE bucket |
| barrier the prefill scan before it overwrites the cell its neighbour read | **correctness fix**: a race that was silent only because of the launch shape | — |

### 3.5 Small-T and projections

Four rows per CTA in the K=2048 decode projections; a direct decode out-projection path for the
35B, then sixteen rows per CTA for it, then making it the default; explicit launch bounds for the
small-T MoE down kernel; one CTA per column for the small-T router reduction; stop sending
two-column rounds to the one-row D4 schedule; widen the Q5 routed-down two-row window to eleven
tokens.

### 3.6 Host side

| change | mechanism | gain |
|---|---|---:|
| memoise the BPE merge result per distinct word within an encode call | repeated words in one prompt were re-merged | render 16 ms → 5 ms |
| skip NFC normalisation for pure ASCII | the scan is cheaper than the transform it usually skips | render |
| give 35B-A3B verify routes of different attention paths distinct graph execution profiles | **correctness fix**: two routes shared one profile, so the graph was rebuilt on every alternation between them | — |
| say what changed when a decode-graph update fails | diagnostics; the previous message did not name the field | — |

### 3.7 Qwen3.8-27B NVFP4

This branch is worth as much to the NVFP4 profile as to the MoE one, and the two commits that
name it are only part of that. Measured on `qwen3_8_27b_nvfp4`, chunk 4096, int8 KV, two
alternating rounds, medians:

| | `dev` 59febed2 | this branch | |
|---|---:|---:|---:|
| prefill 32K | 7 187 t/s | **7 749 t/s** | **+7.8%** |
| prefill 128K | 4 055 t/s | **4 484 t/s** | **+10.6%** |
| decode 32K / 128K | 81.1 / 71.0 t/s | 81.6 / 71.4 t/s | +0.6% / +0.5% |

Decode barely moves and that is expected: on this profile it already runs near the memory
roofline, and the decode work in this branch is MoE and W8, neither of which exists here. The
whole gain is prefill, and it grows with context.

Four commits carry it:

* **register the fused TMA SwiGLU for every chunk width it accepts** — it was registered only at
  exactly `T == 1024`, so any wider chunk fell back to materialising a `34816×T` bf16 arena and
  running a separate `silu_mul`. This is the single largest item on this model.
* **walk the TMA grid in token groups** so an expert's weights stay in L2 across the tiles that
  use them.
* **PV tiles in fp16** (§3.2) — +0.4% at 32K but **+6.0% at 128K**; its value grows with length,
  and it needs int8 KV, since the fp16 accumulation lives in the int8 kernel.
* **the GDN prefill convolution written straight into q/k/v** (§3.3) — the main contributor on
  short prompts, since 48 of the 64 layers here are GDN.

Plus a fifteen-token MTP draft window registered for Qwen3.6/3.8-27B.

**Where not to look next, so it costs you nothing to know.** The fused NVFP4 SwiGLU is 16.5% of
prefill and reaches 975 TFLOP/s against a 1945 TFLOP/s tier, which invites the conclusion that
the kernel is slow. It is not: marginal-cost ablations — each component issued twice, output held
byte-identical — put the TMA load path at a 1.79× time multiplier against 1.54× for the MMA and
1.06× for `ldmatrix`. The critical path is memory and it is saturated; during its own half of the
time the kernel issues at exactly the tier. Bigger tiles cannot fix it either: `BlockN=256` needs
256 accumulator registers against a 232 ceiling, and `BlockM>256` is outside what the TMA box
allows.

---

## 4. Ranked by payoff

If you only want the top of the list, this is the order we would take it in:

| rank | change | gain | lines | independent? |
|---:|---|---|---:|---|
| 1 | prefill chunk 1024 → 4096 (**a flag, not a patch**) | +19…21% prefill | 0 | yes |
| 2 | let the grouped route slice wider than 4096 | unlocks chunk 8192 | ~24 | yes |
| 3 | sigmoid gate inside the attention epilogue | +2.4…3.8% decode | ~79 | yes |
| 4 | prefetch next projection weights from the D4 tail | +2.5% decode | ~96 | yes |
| 5 | prefetch shared-expert down weights from the D1 tail | +1.9% decode | ~17 | needs 4 |
| 6 | fuse the q/k rmsnorm | +1.7% decode | ~129 | yes |
| 7 | PV tiles in fp16 on consumer arches | +4.3…9.2% attention | ~34 | yes |
| 8 | fold the pre-MoE rmsnorm into D1/D3 | +1.2% decode | ~150 | yes |
| 9 | hoist the shared-expert W8 group scales | +1.2% decode | ~41 | yes |
| 10 | fold the router selection into the last D1 block | +0.7% decode | ~31 | needs 8 |

Items 1 and 2 are the cheapest ratio in the whole document: one flag and one bound.

---

## 5. What depends on what

Almost everything is independent. The two real chains:

* router-selection fold **→** pre-MoE rmsnorm fold **→** the three D-tail prefetches.
  The selection fold assumes the D1 prologue already owns the normalised row.
* the MTP draft head routing **→** the fused q/k rmsnorm op it routes into.

Everything else can be taken in any order or alone. The GDN, attention, prefill-MoE, small-T,
host-side and 27B groups do not touch each other.

---

## 6. If you would rather re-implement than merge

That is a reasonable outcome and we have written this with it in mind — no attribution is being
asked for. Each subsection of §3 is one paragraph per idea, with the mechanism stated in terms of
the invariant it relies on rather than the diff, so it can be rebuilt without reading the patch.

Three things worth keeping whichever way you take them:

1. **The identity gate is what makes any of it safe to land.** Greedy `cmp` at a fixed chunk
   width, run on both a fresh process and a warm one. Several of these changes are bit-exact only
   because a rounding step was preserved exactly — the fused norm reproduces `rmsnorm_warp_bf16x2`'s
   BF16x2 rounding, the fused gate reads the reduced value back through the same BF16 rounding.
   A version that skips either is faster and wrong, and only the gate tells them apart.
2. **Prefetch placement is the result, not the prefetch.** The same instruction after the barrier
   is worth nothing; the gain is in issuing it while another warp is still the long pole.
3. **Two of the commits are correctness fixes independent of performance** — the GDN convolution
   tap carrying an unrounded value, and the prefill scan overwriting a cell its neighbour still
   needed. Those are worth taking even if none of the rest is.

---

## 7. Measured and rejected

So that none of this costs you a second time. All measured on the same machine, same gates.

| tried | result | why it is here |
|---|---|---|
| rope folded into the fused q/k norm kernel | last-bit drift, ≈ −1% MTP throughput via acceptance | the obvious next fusion; it is a trap |
| the gate fusion in the MTP attention epilogue as well | bit-identical but **−0.04% MTP3** | two vertices per step against ten already-fused full-attention layers |
| explicit launch bounds on the T=1 MoE down kernel | 22 056 → 21 397 ns in isolation, no end-to-end effect | reverted; the commit pair is not in this branch |
| `cp.async.bulk … .multicast::cluster` for shared expert tiles | correct data, **≈ 300× slower**; a single-CTA cluster with `mask=1` still costs 88.6 ms against 0.148 ms | the instruction assembles and runs on sm_120a but there is no hardware path under it |
| L2 persistence window (`accessPolicyWindow`, 60 MB available) | +1.0% sequential, +1.9% interleaved | 96 MB of L2 already holds a working set that size |
| 256-bit global loads (`ld.global.v8.b32`) | no bandwidth change; DRAM already at 1 700 GB/s = 94.7% of peak | it saves issue slots, not bandwidth |
| int8 for the W8 row-split prefill GEMM | +7.3% **only** if activation quantization is hoisted out of the kernel; quantizing in-kernel makes it −2.9% | each row block re-quantizes the same activations |
| 2:4 structured sparsity | the tier is real and measured at 15.6× the bf16 tier on this card | needs a pruned checkpoint; it is a model change, not a kernel change |

---

## 8. Contracts, portability, risk

* **No CLI, serving, artifact, tokenizer or documentation contract changes.** No new flags, no
  schema bump, no default moved. The one documentation edit describes existing behaviour.
* **Bit-exactness is the acceptance criterion, not a side effect.** Where a change could not be
  made bit-exact it was either dropped (rope) or split so that the exact part lands and the
  inexact part does not (the fused q/k norm is exactly this: the follow-up commit takes rope back
  out to restore identity, and pays 10.1 µs/token to do it).
* **Architecture-specific:** the fp16 PV accumulation is gated to consumer arches, where
  `m16n8k16.f16.f16.f16.f16` runs at twice the f32-accumulate rate. On parts where it does not,
  the existing path is taken.
* **Model-specific:** the two NVFP4 commits and the fifteen-token draft window apply to
  Qwen3.6/3.8-27B; the direct out-projection path is registered for the 35B.
* **Measured on one machine**, an RTX 5090 at a 500 W cap with CUDA 13.1. The relative numbers
  should hold on any sm_120 part; the absolute ones will not.
* **Known unrelated failure:** `ninfer_qwen3_6_27b_prefix_real_test` fails on this branch — and
  fails identically on clean `59febed2` (`path=0 reused=0 prompt=280`, same duration). It looks
  like it belongs to `fix(cache): retain prefixes across rewritten suffixes` rather than to
  anything here, and is flagged so it is not attributed to this branch.

### Known issue: the GDN post-norm tickets sit in a module-global symbol

`recurrent.cuh` declares `__device__ unsigned int g_gdn_post_norm_ticket[64 * 8]`, indexed only by
value head and batch row. The tail election is safe because every slot receives exactly
`gridDim.z` increments per launch and the wrapping `atomicInc` returns it to zero on its own — but
that invariant is **per launch, not per Program**. Two Programs running these kernels at once
would interleave their increments, and one could elect a final block before its own tiles had
finished, so the fused norm would read incomplete output. It is not reachable in the single-Program
configuration everything here was measured in, which is exactly why it is written down rather than
left to be discovered: it is a latent violation of the ownership boundary, not a bug I can point at.

The fix belongs where Program-lifetime device scratch lives, and that is a decision for this
repository rather than for us. What is already known about its shape, so nobody has to rediscover it:

* the kernel side is one line at each of the two `atomicInc` sites once a pointer reaches the
  access structs, which already carry the other `post_norm_*` fields;
* **`TextContext` cannot own it.** A `DeviceBuffer` member zeroed in its constructor compiles and
  removes the symbol, then fails at run time with `cudaMalloc failed:
  cudaErrorStreamCaptureUnsupported` — that constructor runs while the stream is capturing;
* `LinearAttentionStatePool` and `GdnReplayRecords` are non-owning views over a `DeviceSpan`
  somebody upstream allocates, so either one means extending a state layout;
* the workspace arena is reused inside a step, which would break the zero-on-entry invariant the
  tickets depend on.

That leaves allocation at Program construction, outside capture, handed down — landing squarely on
the ownership surface `refactor(runtime): make program the resource authority` just established.
Say where it should live and it is a small change; the three post-mixer-norm commits that use the
tickets are also separable on request, and are worth about 2% of decode between them.

This supersedes #67 and #69, which were built against the previous `master` and no longer apply:
`dev` deleted `include/ninfer/ops/gqa_attention.h`, which #67 modifies. The versions in this
branch are re-derived on the current tree, and the attention one is materially stronger than the
one in #67 — it fuses the batched branch, which the earlier version did not.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -171,7 +171,7 @@ measured recommendation rather than a semantic limit.
 | `--device N` | CUDA device index | `0` |
 | `--kv-dtype bf16\|int8\|fp8` | KV-cache storage | `bf16` |
 | `--spec mtp\|dflash` | speculative backend | off |
-| `--draft-tokens N` | MTP `1..5`; DFlash `1..15` | unset |
+| `--draft-tokens N` | MTP `1..15` where the target registers that window, otherwise `1..5`; DFlash `1..15` | unset |
 | `--lm-head-draft` | optimized proposal head | off |
 | `--vision` | enable image/video input and load Vision GPU allocations | off |
 | `--no-cuda-graph` | disable CUDA Graph decode | graphs on |
--- a/docs/serving.md
+++ b/docs/serving.md
@@ -514,7 +514,7 @@ curl http://127.0.0.1:8080/v1/models \
 | `--response-store-max-mib N` | total local Response envelope/Item/context budget | `256` |
 | `--kv-dtype bf16\|int8\|fp8` | KV-cache storage | `bf16` |
 | `--spec mtp\|dflash` | speculative backend | off |
-| `--draft-tokens N` | MTP `1..5`; DFlash `1..15` | unset |
+| `--draft-tokens N` | MTP `1..15` where the target registers that window, otherwise `1..5`; DFlash `1..15` | unset |
 | `--lm-head-draft` | optimized proposal head | off |
 | `--default-max-tokens N` | output limit when omitted by a request | `8192` |
 | `--default-thinking-budget N` | positive thinking cap inherited by thinking-enabled requests | unset |
--- a/include/ninfer/ops/causal_conv1d_silu.h
+++ b/include/ninfer/ops/causal_conv1d_silu.h
@@ -30,6 +30,15 @@ void causal_conv1d_silu(const Tensor& x, const Tensor& weight, Tensor& conv_stat
 void causal_conv1d_silu(const Tensor& x, const Tensor& weight, const Tensor& conv_state_in,
                         Tensor& conv_state_out, Tensor& out, cudaStream_t stream);
 
+// Split-output form. Same result as the distinct-state form followed by copying the three
+// leading channel ranges of `out` into out_q/out_k/out_v, without materialising the packed
+// buffer. out_q.ne[0] + out_k.ne[0] + out_v.ne[0] must equal x.ne[0], and every destination
+// shares x's column count. `scratch` is a packed [C,T] buffer used only by the short-sequence
+// fallback, where no split kernel exists.
+void causal_conv1d_silu_split(const Tensor& x, const Tensor& weight, const Tensor& conv_state_in,
+                              Tensor& conv_state_out, Tensor& out_q, Tensor& out_k, Tensor& out_v,
+                              Tensor& scratch, cudaStream_t stream);
+
 /**
  * Snapshot form for B independent sequences. `x` and `out` are contiguous BF16 [C,W,B],
  * `conv_states` is contiguous BF16 [C,3,Slots], and `initial_state_slots` and
--- a/include/ninfer/ops/gated_delta_net.h
+++ b/include/ninfer/ops/gated_delta_net.h
@@ -70,12 +70,16 @@ void gated_delta_net(const Tensor& q, const Tensor& k, const Tensor& v, const Te
  * destination_state_slots[b]; selectors may be equal for an in-place row. Destination slots are
  * distinct across active rows. This form uses no arena allocation.
  */
+// `prefetch_data`/`prefetch_bytes` name a weight span the caller knows the next consumer will
+// stream. The tail issues fire-and-forget L2 prefetches over as much of it as the grid covers.
+// Purely a cache hint: it reads nothing the kernel uses and has no numeric effect.
 void gated_delta_net_batch_update(const Tensor& q, const Tensor& k, const Tensor& v,
                                   const Tensor& g, const Tensor& beta, float scale,
                                   bool normalize_qk, Tensor& ssm_states,
                                   const Tensor& source_state_slots,
                                   const Tensor& destination_state_slots, Tensor& out,
-                                  cudaStream_t stream);
+                                  cudaStream_t stream, const void* prefetch_data = nullptr,
+                                  std::size_t prefetch_bytes = 0);
 
 /**
  * Op: gated_delta_net_replay_record
--- a/include/ninfer/ops/gated_delta_net.h
+++ b/include/ninfer/ops/gated_delta_net.h
@@ -92,6 +92,22 @@ void gated_delta_net_batch_update(const Tensor& q, const Tensor& k, const Tensor
  * unchanged, while the invalid out suffix is exact BF16 zero. Inputs, state, records, and out are
  * pairwise non-overlapping.
  */
+// Registration for the gated post-mixer RMS-norm folded into a recurrent kernel's tail. Set it
+// immediately before the call it belongs to; the launcher consumes and clears it, so a launch that
+// does not read it cannot inherit a stale span. The tail keeps one completion ticket per
+// (value head, batch row), which bounds both dimensions.
+inline constexpr std::int32_t kGdnPostNormMaxHeads = 64;
+inline constexpr std::int32_t kGdnPostNormMaxBatch = 8;
+
+struct GdnPostNormSpan {
+    const void* gate_z = nullptr;
+    const void* weight = nullptr;
+    void* out          = nullptr;
+    float eps          = 0.0F;
+};
+void set_gdn_post_norm(GdnPostNormSpan span);
+GdnPostNormSpan take_gdn_post_norm();
+
 void gated_delta_net_replay_record(const Tensor& q, const Tensor& k, const Tensor& v,
                                    const Tensor& g, const Tensor& beta, float scale,
                                    const Tensor& ssm_states, const Tensor& valid_columns,
--- a/include/ninfer/ops/rope.h
+++ b/include/ninfer/ops/rope.h
@@ -40,4 +40,13 @@ void rope(const Tensor& positions, int rotary_dim, float theta, Tensor& q, Tenso
 // from x; Q versus K role does not change the transformation.
 void rope(const Tensor& positions, int rotary_dim, float theta, Tensor& x, cudaStream_t stream);
 
+/**
+ * Fused q/k RMS-norm (unit-offset weights) + rotary. For the Text 16Q/2K, D=256,
+ * rotary 64 geometry this runs one kernel; any other geometry falls back to
+ * rmsnorm+rmsnorm+rope with identical results.
+ */
+void qk_norm_rope(const Tensor& positions, int rotary_dim, float theta, const Tensor& q_in,
+                  const Tensor& q_weight, Tensor& q_out, const Tensor& k_in,
+                  const Tensor& k_weight, Tensor& k_out, float eps, cudaStream_t stream);
+
 } // namespace ninfer::ops
--- a/include/ninfer/ops/softmax_attention.h
+++ b/include/ninfer/ops/softmax_attention.h
@@ -128,12 +128,16 @@ void packed_softmax_attention(const Tensor& q, const Tensor& k, const Tensor& v,
  * non-overlapping. The Op overwrites every addressed cache row but owns no cache allocation,
  * frontier, request identity, or commit authority.
  */
+// An optional `gate` asks the Op to finish with out *= sigmoid(gate), shaped like out. Where the
+// route allows it the multiply is folded into the reduce epilogue; everywhere else the Op applies
+// the standalone elementwise kernel itself. Either way the result is bit-identical to calling
+// sigmoid_mul on the ungated output, so the caller need not know which route it landed on.
 void causal_softmax_attention(const Tensor& q, const Tensor& k, const Tensor& v,
                               const Tensor& positions, const Tensor& valid_columns,
                               const Tensor& kv_table_rows, AttentionHeadGeometry geometry,
                               float scale, PagedKVBatchLayerView cache,
                               CausalAttentionExecutionEnvelope envelope, WorkspaceArena& workspace,
-                              Tensor& out, cudaStream_t stream);
+                              Tensor& out, cudaStream_t stream, const Tensor* gate = nullptr);
 
 /**
  * Read-only single-sequence causal attention over an already populated cache.
--- a/include/ninfer/ops/sparse_moe.h
+++ b/include/ninfer/ops/sparse_moe.h
@@ -22,6 +22,19 @@ enum class SparseMoeEpilogue : std::uint8_t {
     AddResidual,
 };
 
+struct WeightPrefetchSpan {
+    const void* data  = nullptr;
+    std::size_t bytes = 0;
+};
+
+/**
+ * Registers the weight span the NEXT decode-step consumer will stream.
+ * The next sparse_moe decode launch issues fire-and-forget L2 prefetches for it
+ * from the D4 epilogue and clears the registration. Purely a cache hint recorded
+ * at capture time; no numeric effect.
+ */
+void set_next_weight_prefetch(WeightPrefetchSpan span);
+
 /**
  * Returns the transient capacity required by SparseMoe for every T in the inclusive
  * [min_tokens,max_tokens] interval. The routed QTypes are the fixed implementation profile.
--- a/include/ninfer/ops/sparse_moe.h
+++ b/include/ninfer/ops/sparse_moe.h
@@ -73,6 +73,7 @@ void set_next_weight_prefetch(WeightPrefetchSpan span);
  * graph-stable transient storage and carries no state beyond the call.
  */
 void sparse_moe(const Tensor& x, const SparseMoeWeights& weights, SparseMoeEpilogue epilogue,
-                Tensor& destination, WorkspaceArena& workspace, cudaStream_t stream);
+                Tensor& destination, WorkspaceArena& workspace, cudaStream_t stream,
+                const Tensor* pre_norm_weight = nullptr, float pre_norm_eps = 0.0f);
 
 } // namespace ninfer::ops
--- a/src/core/decode_graph.cpp
+++ b/src/core/decode_graph.cpp
@@ -4,6 +4,7 @@
 
 #include <cstdio>
 #include <stdexcept>
+#include <cstddef>
 #include <string>
 
 namespace ninfer {
@@ -15,6 +16,8 @@ void log_cuda_error(const char* op, cudaError_t err) noexcept {
                      cudaGetErrorString(err));
     }
 }
+
+std::size_t g_last_instantiated_nodes = 0;
 
 void destroy_graph_exec(cudaGraphExec_t& exec) noexcept {
     if (exec != nullptr) {
@@ -105,6 +108,12 @@ void DecodeGraphExecutable::instantiate(const DecodeGraphDefinition& definition)
     reset();
 
     cudaGraphExec_t exec  = nullptr;
+    {
+        std::size_t captured = 0;
+        if (cudaGraphGetNodes(definition.graph_, nullptr, &captured) == cudaSuccess) {
+            g_last_instantiated_nodes = captured;
+        }
+    }
     const cudaError_t err = cudaGraphInstantiate(&exec, definition.graph_, 0);
     if (err != cudaSuccess) {
         destroy_graph_exec(exec);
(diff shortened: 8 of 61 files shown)


MichaelDementii:
Thanks — the first one was a real defect and it is fixed; the second is a fair reading of an
invariant I should not have leaned on, and I would like to agree on the shape of the fix before
changing it.

**P1, `NINFER_MTP_DRAFT_DEPTH` — fixed, and it was not alone.** The cap was profiling
instrumentation that should never have reached the branch, and your reading of the consequence is
right: the loop stops early while the round still publishes `k` extents, so the omitted suffix is
verified as real draft tokens. Auditing for the same mistake found six environment switches in
total, all of them sweep knobs of mine:

| removed | was |
|---|---|
| `NINFER_MTP_DRAFT_DEPTH` | draft-depth cap; now the loop runs to `k` |
| `NINFER_MOE_BN` | pinned the prefill MoE tile width; now derived from the chunk, as intended |
| `NINFER_MOE_D3`, `NINFER_MOE_D4`, `NINFER_MOE_D4_MAIN`, `NINFER_MOE_D4_MTP` | small-T schedule overrides; the resolver's chosen point now stands alone |

A seventh, `NINFER_OUTPROJ_DIRECT`, was removed before the branch was opened for the same reason.
In every case the behaviour with the variable unset is preserved exactly, which is also the
configuration every number in the description was measured under. Re-verified after the change:
`ctest` 94/94, greedy output still byte-identical to `dev` on all fixtures, decode and MTP-3
unchanged within run-to-run drift. Force-pushed as `c735909b`.

**P1, the GDN completion tickets — I agree the symbol is in the wrong place.** The array is a
module-global `__device__` symbol indexed only by value head and batch row, so two Programs
executing these kernels concurrently would share it, and the self-resetting `atomicInc` invariant
that makes the tail election safe holds per launch, not per Program. It is not reachable in the
configuration I measured, which is exactly why it should not stay: it is a latent violation of
the ownership boundary rather than a bug I can point at.

The fix I would make is the one you describe — move the tickets into Program-owned workspace and
pass the pointer into the two kernels that take them (`recurrent.cuh`, the two `atomicInc` sites),
sized like the rest of the GDN workspace. Before I do that I would rather ask, since it touches
the ownership surface the recent `refactor(runtime): make program the resource authority` just
established: is the GDN workspace allocation the right home for this, or would you prefer these
tickets live in Program state next to the other graph-stable device scratch? I will follow
whichever you name. If you would rather not carry the three post-mixer-norm commits at all while
this is open, they are separable — they are worth about 2% of decode between them, and the rest
of the branch does not depend on them.

MichaelDementii:
Following up on the ticket ownership with something concrete, since it narrows the answer.

I tried the obvious placement first — a small `DeviceBuffer` owned by `TextContext`, zeroed in its
constructor, published through `GdnPostNormSpan` into both access structs. It compiles and the
global symbol disappears, but it does not run:

```
error: cudaMalloc failed: cudaErrorStreamCaptureUnsupported:
       operation not permitted when stream is capturing
```

`TextContext` is constructed while the stream is capturing, so the storage cannot be allocated
there. The other two candidates I looked at are non-owning views over spans somebody upstream
allocates — `LinearAttentionStatePool` and `GdnReplayRecords` both take a `DeviceSpan backing` —
so putting the tickets in either means extending a state layout, and the workspace arena is
reused within a step, which would break the very invariant the tickets rely on (each slot must
still be zero when the next launch starts).

That leaves the placement you would expect: allocated once at Program construction, outside
capture, and handed down. I have not made that change, because it lands exactly on the ownership
surface `refactor(runtime): make program the resource authority` just established and I would
rather not guess at its shape from the outside.

So, concretely, whichever you prefer:

1. **Tell me where Program-lifetime device scratch should live** and I will thread it through and
   re-verify — the kernel side is a one-line change at each of the two `atomicInc` sites once the
   pointer arrives in the access struct.
2. **Or I drop the three post-mixer-norm commits from this branch**, which removes the global
   symbol entirely and leaves the rest untouched. They are separable and worth roughly 2% of
   decode between them; say the word and I will push without them, and they can come back as
   their own pull request once the placement is settled.

The branch as it stands is unchanged from what you reviewed apart from the environment switches
being gone: build clean, `ctest` 94/94, greedy output still byte-identical to `dev`.

Neroued:
Thanks for the detailed work and measurements. I’m going to close this PR rather than continue iterating on the current branch.
- A 38-commit, 61-file performance stack is not reviewable or maintainable as one change. Individually described commits do not make the combined state independently verifiable.
- Some of the debugging work was valuable. The Sparse-MoE prefill scan race was a real upstream bug and has been fixed independently in fbd04729. The proposed GDN BF16 carry, however, is not an upstream correctness bug: that accumulator is private arithmetic, not a required BF16 state boundary. Problems introduced by this branch, such as the MTP profiling cap and global GDN completion tickets, should also not be counted as existing upstream bugs.
- The default prefill chunk of 1024 is deliberate. The reported workspace grows from about 120 MiB at 1024 to 482 MiB at 4096 and 963 MiB at 8192. That memory comes directly out of the capacity available for KV, cached contexts, and active requests, so the larger-chunk gain is not free and does not justify changing the default.
- Several performance ideas are worth evaluating independently. Aggregate end-to-end numbers over dozens of changes contain too much noise and interaction to attribute a gain. A new PR should contain one mechanism, or one inseparable dependency chain, rebased on current master, with production-shape operator benchmarks, numerical validation, and resource deltas. Cross-kernel fusion and prefetch work should additionally measure the relevant producer-consumer or CUDA Graph step. End-to-end measurements should remain the final confirmation, not the sole evidence.
- Please do not bundle product-contract changes into performance work. Expanding MTP from 5 to 15 drafts changes the CLI domain, state/workspace bounds, and Graph profiles; it requires a separate proposal and workload-level evaluation. Hidden mutable channels such as thread-local setters or module-global device state are also not acceptable ownership.
- If you want to continue, please open new, narrowly scoped PRs based on current master. Attention/norm fusion, direct GDN Q/K/V output, and individual MoE/projection kernel optimizations are reasonable examples of independently reviewable work. The current combined PR will not be merged.

PR #97, opened 2026-08-26
build: cache native compilation and incremental image builds

DuncanBetts:
Upstream image rebuilds discard the entire CMake/Ninja build tree when source files change. This PR retains that tree and adds ccache: editing one C++ file and one CUDA file then runs seven affected build steps instead of all 278. Native builds can enable the same compiler cache through standard CMake launchers.

- Document optional native CMake compiler launchers, incremental rebuilds, cache configuration, statistics, and how to disable the launchers again. The ordinary native build needs neither ccache nor container tooling.
- Retain the CMake/Ninja tree in a locked build cache mount. Synchronize sources by content, remove deleted files, and select separate build directories from installed package versions, the Dockerfile, and CMake scripts. Removed configure flags and changed cached defaults therefore cannot leak into a different configuration. Copy completed binaries out of the cache for the runtime image.
- Install ccache and rsync only in the build stage. Keep ccache in a separate mount with a 20 GiB limit, so compiler outputs can also be reused when Ninja needs to compile again.
- Use the existing Docker workflow. No Compose file, Podman-specific workflow, extra build stage, or Python dependency is introduced.
- Add a standalone build-cache regression check using the real Dockerfile with a tiny CMake fixture. It needs no GPU or model and does not compile or run NInfer.

No project-specific CMake option is needed: CMake already supports compiler launchers for all three languages. The runtime image and inference behavior are unchanged.

### Rebuild comparison: upstream versus this PR

These measurements and recordings are from `c0392a3b`, before the configuration-invalidation follow-up. The timings have not been rerun for the updated cache key; the new regression check verifies incremental compilation behavior, not an updated performance figure.

Both recordings follow the same workload: build the image, add the same harmless `static_assert` to `src/artifact/typed_binding.cpp` and `src/ops/launcher/residual_add.cu`, rebuild, then restore the sources. The baseline is upstream without this PR; the after result uses persistent Ninja and compiler caches at that revision.

| Rebuild after the two-file edit | Upstream without this PR | With this PR |
| --- | ---: | ---: |
| Image rebuild wall time | 127.750 s | 20.090 s |
| Ninja steps | 278 | 7 |
| Compiler invocations | 263 | 2 |

The recorded rebuild is approximately 6.4× faster. These are single-run measurements recorded separately, not a contemporaneous controlled benchmark. The baseline is the original upstream recording; it was not rerun alongside that revision. The times cover image rebuilding, not native-build or inference performance.

The after run started with empty, isolated build and compiler caches. Its initial build made 263 compiler calls; the edited rebuild made only two more. Ninja skipped all 261 unchanged compiler inputs, so the cumulative ccache statistics are 265 calls, zero hits, and 265 misses. CUDA device linking still took about 13.4 seconds on the rebuild.

- [Upstream recording, without this PR](https://asciinema.org/a/ZICrComRvwPQGqzU)
- [Pre-review recording with this PR, including five chapter markers](https://asciinema.org/a/jAaiTxConG6A7GRi)

### Original image validation (`c0392a3b`)

Using Ubuntu 24.04, CMake 3.28.3, CUDA 13.1.115, and ccache 4.9.1:

- The revised Dockerfile built both product binaries and the runtime image successfully (278 Ninja steps; all 263 compiler calls cacheable).
- Restoring the two source files with timestamps from 2000 correctly rebuilt both objects (two ccache hits).
- Changing only non-code build context reported `ninja: no work to do`, with no new compiler-cache calls.
- After clearing all three launchers and moving ccache out of PATH in a disposable build container, the C++ artifact-binding and CUDA residual-add objects compiled successfully.
- Both runtime binaries passed `--help` with NVIDIA driver libraries exposed; the runtime image contains neither ccache nor rsync.
- `git diff --check` passed.

Validation used Podman's Dockerfile support because Docker and a native host CMake/CUDA toolchain are not installed here. Docker BuildKit and a bare-host build were not executed; the native instructions use standard CMake settings and introduce no container dependency.

### Configuration-invalidation regression

The new check is:

```bash
bash tests/test_docker_build_cache.sh docker
```

It runs the actual build stage against a tiny C++/CMake fixture, substitutes isolated cache IDs, and changes a non-code context marker before each build so an image-layer hit cannot hide stale configuration. Compile-time assertions check the active compiler flag and cached default; the fixture binaries are not executed.

Validated with `bash tests/test_docker_build_cache.sh podman`, using Podman 5.8.6, the Ubuntu 24.04 build image, CMake 3.28.3, Ninja 1.11.1, GCC 13.3.0, and ccache 4.9.1:

- All nine build scenarios passed: cold build; flag added and removed; non-code edit; one-file source edit; cached default changed and restored; environment flag added and removed.
- Flag removal also restores the expected-value header with a timestamp from 2000.
- The non-code edit reported `ninja: no work to do`; the source edit rebuilt exactly one translation unit.
- As a negative control, restoring the previous package-only cache key made the flag-removal case fail with `static assertion failed: stale compiler flags`. This confirms the regression check detects the reported defect.
- `bash -n tests/test_docker_build_cache.sh` and `git diff --check` passed.

Docker BuildKit remains untested because Docker is unavailable here. The new validation covers configuration and incremental-build behavior with a small fixture, not a fresh CUDA/product build or inference run. The original full-image results above are retained as historical evidence, not claimed as rerun.

### Reviewer notes

- Only the build-directory key changes in the production build step; the standard launchers, source synchronization, lock, and runtime-image publication remain intact.
- Hashing Dockerfile and CMake-script contents is deliberately conservative: even a comment-only edit to those files selects a different configuration. Ordinary C++/CUDA source edits retain the Ninja tree, and ccache remains shared across configurations.
- The fixed, locked build cache serializes concurrent builds on the same builder, including different checkouts. Branch switches update the shared source mirror.
- The 20 GiB limit applies only to ccache. Separate Ninja directories accumulate for different configurations and remain until the build cache is removed or reclaimed by the builder. This cost and the serialization tradeoff are now documented in the README.
- rsync is not part of the final image.

### References

- [CMake compiler launchers](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LAUNCHER.html)
- [ccache manual](https://ccache.dev/manual/latest.html)
- [Docker build cache mounts](https://docs.docker.com/build/cache/optimize/#use-cache-mounts)

Changes made using AI. Thanks for creating Ninfer Neroued, I hope you find these build time savings useful in future work, you can take it as a token of my appreciation :).


DuncanBetts:
### Review findings and resolution

The review found that removing a compiler flag from the Dockerfile could leave it active in the persistent CMake cache, making the binary depend on build history.

Fixed in [9d89a654](https://github.com/Neroued/ninfer/pull/97/commits/9d89a65411594fcdae9e09773a7ff20c644b62ab): the build-directory key now includes the Dockerfile and CMake scripts alongside installed package versions. Ordinary source edits retain the Ninja tree; ccache remains shared across configurations. Both optimizations are retained.

**Verification:** all nine scenarios in `bash tests/test_docker_build_cache.sh podman` passed using the real Dockerfile with a tiny CMake fixture and isolated caches. They cover flag removal, cached-default and environment changes, old timestamps, and incremental rebuilding. Restoring the old key reproduced the stale-flags failure. Non-code edits did no compilation; a one-file edit rebuilt exactly one translation unit. Shell syntax and `git diff --check` also passed. Toolchain details are in the PR description.

**Remaining tradeoffs:** builds sharing the cache serialize, branch switches update the shared source mirror, and the 20 GiB limit covers ccache—not accumulated Ninja directories. These costs and the configuration-maintenance responsibility are documented in the consolidated README guidance.

The earlier 6.4× image-rebuild result primarily reflects Ninja skipping compilation, not ccache hits; it has not been remeasured after the fix and is not an inference-performance claim.

Docker BuildKit remains untested because Docker is unavailable here. This follow-up did not run NInfer or rebuild the full CUDA/product image. The PR is now ready for review.

DuncanBetts:
Addressed the review in [9d89a654](https://github.com/Neroued/ninfer/pull/97/commits/9d89a65411594fcdae9e09773a7ff20c644b62ab).

- **Configuration invalidation:** the build-directory key now hashes the Dockerfile and all `CMakeLists.txt`/`*.cmake` files as well as installed package versions. Different configure recipes/defaults no longer reuse one CMake cache. Ordinary source edits still retain the incremental Ninja tree, and ccache is shared across configurations.
- **Regression coverage:** added `tests/test_docker_build_cache.sh`, which exercises the real Dockerfile with a tiny CMake fixture and isolated caches. All nine build scenarios passed under Podman 5.8.6 / Ubuntu 24.04 / CMake 3.28.3 / Ninja 1.11.1 / GCC 13.3.0 / ccache 4.9.1. The tests cover flag addition/removal, cached-default changes/restoration, environment flags, old header timestamps, non-code edits, and a one-file incremental rebuild.
- **Defect reproduced:** as a negative control, switching back to the previous package-only key made flag removal fail with `static assertion failed: stale compiler flags`. The updated key passes. The non-code case does no compilation, and the one-file edit rebuilds exactly one translation unit.
- **Tradeoffs documented:** the README now explains cross-checkout serialization, the shared source mirror, and that the 20 GiB limit covers ccache only—not accumulated Ninja build directories. The lock remains intentionally unchanged; this update does not add a cache-management framework.
- **PR description clarified:** added reviewer notes and separated the original image/timing evidence at `c0392a3b` from this follow-up's regression results. The 6.4× timing has not been remeasured.

Commands run:

```bash
bash -n tests/test_docker_build_cache.sh
bash tests/test_docker_build_cache.sh podman
git diff --check
```

Docker BuildKit could not be checked because Docker is unavailable here. The same regression script can be run with `docker` instead of `podman`. No NInfer inference or fresh full CUDA/product build was run for this follow-up. The PR remains a draft.

Issue #38, opened 2026-08-17
Official Qwen3.8-27B NVFP4 artifact is much slower than older community version?


s1lverkin:
Hi,

I am testing the official Qwen3.8-27B NVFP4 model:

https://huggingface.co/neroued/Qwen3.8-27B-nvfp4-NInfer

Before this, I was using the older community artifact:

https://huggingface.co/Ostfralla/Qwen3.8-27B-NVFP4-NInfer

I expected the official version to be similar or faster, but in my tests it is quite much slower and also returns more invalid structured responses.

## Test

I tested both versions on an RTX 5090 with:

- The same 250 prompts
- The same client and sampling settings
- 16k context
- INT8 KV cache
- MTP with 3 draft tokens
- Maximum server concurrency of 8

### Time to first token

- Older community artifact: **50.5 ms**
- Official artifact: **276.4 ms**

The older version is around **5.5x faster** here.

### Concurrency 2

- Prefill:
  - Older: **2482 tok/s**
  - Official: **2062 tok/s**

- Decode:
  - Older: **290.7 tok/s**
  - Official: **234.0 tok/s**

- Valid structured responses:
  - Older: **226/250**
  - Official: **203/250**

### Concurrency 8

- Prefill:
  - Older: **5243 tok/s**
  - Official: **3930 tok/s**

- Decode:
  - Older: **602.1 tok/s**
  - Official: **441.4 tok/s**

- Valid structured responses:
  - Older: **222/250**
  - Official: **211/250**

“Valid” only means that the response passed my expected JSON/schema validation. It is not a semantic quality evaluation.

## Runtime difference

One important thing is that I could not run both artifacts on exactly the same NInfer runtime.

The older artifact fails on the current official runtime with this error:

`tensor descriptor does not match target contract: text/token_embedding`

Because of this, I used:

- Current official NInfer runtime for the official artifact
- Older patched NInfer runtime required by the community artifact

I aligned all other settings as much as possible.

## Question

Is this performance difference expected?

Maybe the official artifact uses a different FP8/NVFP4 layout which gives better accuracy but lower performance? Or maybe I am missing some recommended runtime or server option?

Neroued:
A few points here:

The current Qwen3.8 nvfp4 artifact is about 20 GB, and a decode step reads about 17.8 GB of weights. The community artifact uses the same weight format as the previous Qwen3.6 nvfp4 artifact, where it's about 14.3 GB. So this already explains roughly a 20% decode speed difference.

[...]

PR #107, opened 2026-08-28
fix(qwen3.8): wire-format detect nvfp4 artifact profile

koloved:
Qwen3.8-27B NVFP4 ships in two wire formats: the upstream FP8 profile (row-scaled FP8 vocabulary/attention endpoints, 21 GB artifact) and the community W8+NVFP4 profile (W8 vocabulary endpoints, Qwen3.6 NVFP4 layer layout, 18.3 GB artifact). Resolve the weights profile from the artifact's text/token_embedding descriptor instead of always selecting the FP8 profile.

The W8+NVFP4 profile comes from Ostfralla/Qwen3.8-27B-NVFP4-NInfer (https://huggingface.co/Ostfralla/Qwen3.8-27B-NVFP4-NInfer) and runs significantly faster on an RTX 5090 while leaving more VRAM, so it supports a larger context (262144 tokens at full speed) than the 21 GB upstream FP8 artifact, which cannot reach those speeds on the same hardware.

Existing artifacts are unaffected: the upstream FP8 model still resolves to the FP8 profile (Qwen38Nvfp4) because its token_embedding is row-scaled FP8; only W8-embedded community artifacts are routed to Qwen36Nvfp4.

### DIFF
--- a/src/targets/qwen3_6_27b/export/ninfer/targets/qwen3_6_27b/package.h
+++ b/src/targets/qwen3_6_27b/export/ninfer/targets/qwen3_6_27b/package.h
@@ -15,6 +15,7 @@ struct DeviceContext;
 
 namespace artifact {
 class Binder;
+class Reader;
 class MaterializedArtifact;
 struct ArtifactIdentity;
 struct MaterializationPlan;
@@ -121,7 +122,8 @@ struct Package {
     using Program                    = qwen3_6::Program<detail::Variant>;
 
     [[nodiscard]] static ModelSamplingDefaults sampling_defaults(std::string_view model);
-    [[nodiscard]] static WeightsProfile resolve_weights(const artifact::ArtifactIdentity& identity);
+    [[nodiscard]] static WeightsProfile
+    resolve_weights(const artifact::ArtifactIdentity& identity, const artifact::Reader& reader);
     [[nodiscard]] static LoadPlan plan_load(artifact::Binder& binder, const EngineOptions& options,
                                             WeightsProfile weights_profile);
     [[nodiscard]] static std::unique_ptr<LoadedModel>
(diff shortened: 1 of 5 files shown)


stormgraser-ux:
Confirming this from the other side — I've been carrying a local patch doing the same thing, and #107's predicate matches it on both artifacts I have here. Read straight from the container headers on an RTX 5090:

| artifact | bytes | `text/token_embedding` | resolves to |
|---|---|---|---|
| `Ostfralla/Qwen3.8-27B-NVFP4-NInfer` | 18,324,067,840 | `W8G32_F16S` (`row-split-k128-v1`) | `Qwen36Nvfp4` ✓ |
| `huihui-ai/Qwen3.8-27B-abliterated-NVFP4` | 21,492,695,040 | `FP8_E4M3FN_ROW_BF16S` (`row-scale-v1`) | `Qwen38Nvfp4` ✓ |

The second is worth noting: it's a third-party abliterated build that happens to use the official object layout, declares the same `qwen3.8-27b/nvfp4` identity, and still resolves correctly — so the descriptor is discriminating layout rather than provenance.

Without this the Ostfralla artifact dies inside materialisation with a byte-count mismatch naming neither the artifact nor the profile. It reads as a corrupt download, which cost me a re-download before I worked out what it was.

One suggestion: the current code falls through to `Qwen38Nvfp4` when `text/token_embedding` is absent, isn't a tensor, or carries a third format — which reinstates that same unnamed failure for any future artifact. Throwing there, naming both expected formats, makes it a one-line diagnosis.

Running the Ostfralla profile at 262,144 context with vision on the 5090 as a daily driver; happy to test a revision.

knoopx:
i've been also patching my own fork/running `Ostfralla/Qwen3.8-27B-NVFP4-NInfer` for some time, if i recall properly prefill speeds are +10% with this layout over official one and i also verified the conversion carries over and improves `ThinkingCap-Qwen3.6-27B`, there was an open issue regarding this but it was closed https://github.com/Neroued/ninfer/issues/38

PR #122, opened 2026-08-31
fix(runtime): bound pressure planner committed targets

### LINKED ISSUE #121: Pressure planner target accounting can terminate the inference engine
## Description

NInfer's resource-pressure planner can exceed the Qwen target arena even though the planner's configured search budget has not been reported as exhausted.

When this occurs, the active request fails with:

```text
prepared pressure expansion exceeds the target arena
```

The exception reaches the inference worker's fatal failure boundary. The HTTP server remains running, but the inference engine becomes unavailable and all subsequent requests return HTTP 503 until the service is restarted.

This creates a persistent service outage from an ordinary bounded resource-planning condition.

## User-visible impact

A request begins normally and then fails during resource planning. Every later inference request is rejected with:

```text
status=503
code=service_unavailable
message=inference engine is unavailable
```

The container and basic HTTP process can still appear healthy, so a liveness check that does not perform inference may not detect the failure.

## Observed failure sequence

The same sequence was observed twice:

```text
2026-08-30 07:01:47  request 844 submitted
2026-08-30 07:01:49  request 844 error: prepared pressure expansion exceeds the target arena
2026-08-30 07:01:50  request 845 rejected: inference engine is unavailable
```

```text
2026-08-30 08:11:03  request 76 submitted
2026-08-30 08:11:05  request 76 error: prepared pressure expansion exceeds the target arena
2026-08-30 08:11:06  request 77 rejected: inference engine is unavailable
```

The triggering requests were ordinary streaming Responses requests with five messages, 23 tools, no tool-call history, thinking enabled, and a 32,768-token output allowance.

## Relevant configuration

The topology that exposed the defect used:

```text
max concurrency: 1
device checkpoint cache slots: 4
host state slots: 16
private continuations: 12
shared prefixes: 4
long anchors per continuation: 2
host KV arena: 384 GiB
main KV capacity: 240,000 tokens
```

MTP speculative decoding was enabled with three draft tokens and the optimized LM-head draft path. However, the deterministic reproduction is entirely host-side and does not use MTP, draft tokens, or LM-head inputs. MTP is therefore incidental to this defect.

## Root cause

The pressure planner uses a fixed optional-target search budget:

```cpp
inline constexpr std::size_t kOptionalTargetCapacity = 4096;
```

The Qwen pressure-planning session reserves a target arena sized as:

```cpp
candidates.size() + 1 + kOptionalTargetCapacity
```

The resource manager bounds planning by the number of targets that have been **assessed**. That accounting does not include all canonical targets already committed to the planner.

One call to `prepare_expansion()` can generate several unique child targets. `commit_expansion()` interns all of those children, but they remain queued and unassessed. Consequently:

```text
committed canonical target count > assessed target count
```

The planner can therefore reach its next assessment while still appearing to be below the 4,096 assessed-target limit, even though there is insufficient arena capacity to commit that target's complete child expansion.

With many private and shared owners, the planning graph is combinatorial. In a 16-private-owner and four-shared-owner topology, an identity-target expansion can create up to 56 unique children at once.

The existing guard in `commit_expansion()` detects the mismatch:

```cpp
const std::size_t maximum =
    candidates.size() + 1U + planning_detail::kOptionalTargetCapacity;

if (prepared_new_count > maximum - std::min(maximum, targets.size())) {
    throw std::length_error(
        "prepared pressure expansion exceeds the target arena"
    );
}
```

The guard correctly prevents an out-of-bounds commit, but the resource manager should never request a commit that exceeds its declared search budget during an ordinary bounded search.

## Why the engine becomes unavailable

The target-arena exception is not converted into a normal "no resource plan" result. It escapes the request-planning path and reaches `EngineCore::worker_loop()`.

The worker loop treats unexpected exceptions as engine invariant failures, marks the engine failed, and exits the worker. The serving process remains alive, but later requests fail during preparation because the inference engine is no longer available.

This explains the observed split state:

- HTTP process and container remain running
- Basic health endpoint may still return success
- Model and inference routes return HTTP 503
- Restart temporarily restores service

## Expected behavior

The planner's target budget should bound every canonical target admitted to the search, not only targets that have already been assessed.

Before committing an expansion, the resource manager should account for:

- The identity target
- Every canonical target already committed
- Every new canonical child in the proposed expansion

If the complete expansion does not fit within the remaining budget, it should be discarded atomically and treated as normal search-capacity exhaustion.

No partial decisions, targets, or handles should be committed.

A bounded planning search that finds no legal plan should fail only the affected request. It must not permanently disable the inference engine, and a later valid request must remain serviceable.

True stale-handle, generation, or internal-consistency failures should remain fatal rather than being hidden as ordinary capacity exhaustion.

## Suggested fix

Track the number of committed canonical targets against the existing 4,096-target budget.

For each prepared expansion:

1. Read its `new_canonical_count()`.
2. Check whether the current committed-target count plus the complete child set would exceed the search budget.
3. If the full expansion fits, commit it and increment the committed-target count by the number of new canonical children.
4. If it does not fit, discard the expansion without committing decisions, targets, or handles.
5. Continue or terminate the bounded search through the existing normal no-plan path.

The target limit does not need to be increased. Correct accounting makes the resource manager and target arena enforce the same bound.

## Required invariants

The fix should preserve:

- The existing 4,096-target memory bound
- Breadth-first deterministic search ordering
- Deterministic termination
- Stable target ordinals
- Handle and index validity
- Atomic expansion commit or discard
- No partial owner-decision interning
- No partial target publication
- A fatal path for genuine stale-state and consistency errors
- Continued engine availability after ordinary search-capacity exhaustion

## Deterministic reproduction

A resource-manager test can reproduce the failure without a GPU or model artifact by using a fake planner/package with:

- 16 private owners
- Four shared owners
- A full 4,096-target planning budget
- A target arena matching Qwen's `candidate count + identity target + 4096` bound
- Owner successor decisions that create a wide combinatorial expansion graph

### Before the fix

The search deterministically reaches:

```text
prepared pressure expansion exceeds the target arena
```

The planner has committed more canonical targets than the resource manager counted because queued children were excluded from the assessed-target counter.

### After the fix

- The full expansion is discarded when it cannot fit.
- The planning attempt terminates through the normal bounded outcome.
- No target-arena exception is thrown.
- A subsequent valid planning operation using the same planner succeeds.

The final condition demonstrates that capacity exhaustion does not corrupt planner state or leave the service unusable.

## Regression tests

Recommended coverage:

1. A wide expansion with 16 private and four shared owners reaches the 4,096-target budget without exceeding the arena.
2. An expansion whose complete child set cannot fit is discarded atomically.
3. No prepared owner decisions are committed when an expansion is discarded.
4. Stable target ordinals and existing handles remain valid.
5. Search ordering remains deterministic.
6. A later small planning request succeeds after the bounded search is exhausted.
7. Genuine stale expansion handles still throw an invariant error.
8. The same test passes with an arena sized exactly like the Qwen implementation.

SolettaSolaris:
## Summary

- count every committed canonical pressure-planning target against the existing 4,096-target search budget
- discard an expansion atomically when its complete set of new child targets will not fit
- preserve deterministic breadth-first ordering, stable target handles, and the existing memory bound
- add a regression covering 16 private owners, four shared owners, full target-budget pressure, and a subsequent valid planning operation

## Problem

The planner bounded work using the number of targets that had been assessed. A single assessed target can commit several unique children, which remain queued and unassessed. The number of committed canonical targets could therefore exceed the assessed-target count.

With a wide owner topology, the planner could pass its assessment-count check and then attempt to commit more children than the Qwen target arena had remaining capacity for. This raised:

```text
prepared pressure expansion exceeds the target arena
```

The exception could escape the request path, terminate the inference worker, and leave later requests returning HTTP 503 until restart.

## Fix

The resource manager now tracks the identity target and each newly committed canonical child against the same existing target budget. If an expansion's complete child set does not fit, the expansion is discarded before any partial commit.

The 4,096 limit is unchanged. Genuine stale-state and invariant errors remain errors.

## Testing

- driverless resource-manager/planner suite
- resource-manager ASan/UBSan run
- admission-policy driverless test
- KV-capacity driverless test
- `git diff --check`
- formatting checks

The regression deterministically failed before the change and passes afterward. It also verifies that another valid planning operation succeeds after budget exhaustion.

## Production observation

This patch was generated and implemented with AI assistance. It has also been running in a production deployment after review and CI. The original target-arena failure reproduced twice before the patch and has not been observed again since deployment. That is encouraging operational evidence, but it is not presented as a substitute for the deterministic regression test or as proof that every workload is covered.

Closes #121.

### DIFF
--- a/src/runtime/engine/shared_capture_planner.h
+++ b/src/runtime/engine/shared_capture_planner.h
@@ -101,6 +101,7 @@ public:
         queue_.push_back(identity);
         std::optional<Incumbent> incumbent;
         std::uint32_t targets_evaluated = 0;
+        std::uint32_t canonical_targets = 1;
         std::size_t cursor              = 0;
 
         while (cursor < queue_.size() && targets_evaluated < input.target_budget) {
@@ -127,12 +128,13 @@ public:
 
             if (!assessment.expandable || contains(expanded_, target)) { continue; }
             auto prepared                 = session.prepare_expansion(target);
-            const std::uint32_t remaining = input.target_budget - targets_evaluated;
+            const std::uint32_t remaining = input.target_budget - canonical_targets;
             if (prepared.new_canonical_count() > remaining) {
                 session.discard_expansion(std::move(prepared));
                 continue;
             }
             const auto children = session.commit_expansion(std::move(prepared));
+            canonical_targets += children.new_canonical_count;
             expanded_.push_back(target);
             for (const PressureTargetHandle child : children.children) {
                 if (!contains(assessed_, child) && !contains(queue_, child)) {
--- a/tests/test_resource_manager.cpp
+++ b/tests/test_resource_manager.cpp
@@ -954,6 +954,7 @@ public:
     std::uint32_t private_pressure_alternatives     = 1;
     std::uint64_t pressure_action_immediate_ns      = 100'000'000;
     std::uint32_t pressure_action_degradation_units = 1;
+    std::optional<std::uint32_t> pressure_expansion_arena_capacity;
     bool include_cumulative_private_target          = false;
     bool combined_target_cancels_pressure_copy      = false;
     std::optional<std::uint64_t> pressure_target_immediate_ns_override;
@@ -1295,6 +1296,12 @@ FakePressurePlanningSession::commit_expansion(FakePreparedPressureExpansion&& pr
     require(scratch_live_ && prepared.generation == generation_ &&
                 prepared.scratch_generation == scratch_generation_,
             "fake prepared expansion is stale");
+    if (program_->pressure_expansion_arena_capacity) {
+        const std::size_t maximum = *program_->pressure_expansion_arena_capacity;
+        if (prepared.new_count > maximum - std::min(maximum, targets_.size())) {
+            throw std::length_error("prepared pressure expansion exceeds the target arena");
+        }
+    }
     committed_children_.clear();
     std::uint32_t new_count = 0;
     for (Target& child : expansion_scratch_) {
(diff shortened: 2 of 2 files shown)


Neroued:
Sorry, I missed this PR while addressing #121. The implementation is semantically identical to the fix landed in 5e4bf313cb2f8b0603e00bf3b42e7ab3ec6d927a. Thank you for the thorough report and regression coverage, especially the follow up planning check. I’ll close this as already fixed.

PR #126, opened 2026-08-31
Pr1/per layer kv

Astrangemaninhere:
Background
The KV cache previously used a single global storage format for every layer. Attention layers differ sharply in how much quantization error they tolerate, so a uniform format either wastes memory on robust layers (all-INT8) or degrades quality on sensitive ones (all-NVFP4). This PR makes the storage format a per-layer decision.

Changes
Per-layer dtype table: each full-attention layer selects BF16 / INT8 / FP8 independently via --kv-layer-storage or the variant default table.
Layer plane-base prefix sums: mixed-format layers have unequal plane counts and strides; the page geometry now tracks each layer's plane offset.
Scale-plane resolution: FP8 per-layer scale planes are FP16 (not the global format), and per-layer quant_group is resolved into layer views.
Table propagation: layer_kv_dtypes is copied into the sequence candidate so the table actually reaches the cache (previously never applied).
Files
src/targets/qwen3_6/impl/state/decoder_state.cpp (plane geometry, per-layer views)
src/targets/qwen3_6/impl/runtime/layouts_impl.h (candidate propagation)
src/core/paged_kv_cache.h (layer views carry dtype/group/plane base)
Dependency
Foundation for PR2 (cold pool), PR5 (YaRN), and PR6 (NVFP4 tier). Merge first.

### DIFF
--- a/apps/cli/main.cpp
+++ b/apps/cli/main.cpp
@@ -1,4 +1,5 @@
 #include "options.h"
+#include "product/kv_options.h"
 #include "product/load_progress/load_progress.h"
 #include "product/prompt_input/prompt_input.h"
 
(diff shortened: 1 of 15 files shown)


Neroued:
.......

PR #131, opened 2026-08-31
pr6/Nvfp4 cold pool

Astrangemaninhere:
### Background
The integrated NVFP4 work on top of PR1-5: NVFP4 (E2M1 + E4M3 g16 scales, ISO3 V) KV tier, GQA attention ops, a data-driven default layer table, and an entropy-coded cold pool that finally shrinks the device KV footprint.

### Changes

**GQA attention (batch 1)** — `src/ops/kernel/gqa_attention_*`:
- Full GQA attention ops (A1/A2/A3: append+attention / append-only / attention-only) with BF16 / INT8 / FP8 / NVFP4 / ISO3 decode kernels and NVFP4 prefill.
- `gqa_attention` replaces the d256 causal attention path for main text.

**Default layer table** — `src/targets/qwen3_6_27b/impl/variant.cpp`:
- 10/16 layers NVFP4 `{0,1,3,4,6,7,8,9,13,14}`, 6 layers INT8 `{2,5,10,11,12,15}`.
- Measured on 13.3k zh perplexity: **ctx4096 1.3344** vs all-INT8 1.4324; **ctx8192 1.1952** vs 1.2504; long corpus (148k) ctx16384 **12.098** vs 12.099. NVFP4 layers score parity-or-better because the g16 scale grid is finer than INT8 g64.
- The 8/8 neighborhood marginals (not per-layer NMSE/rank metrics) predicted the combo; `analyze_kv.py` metrics were shown not to predict combination outcomes.

**NVFP4-tier cold pool** — entropy rANS slots:
- NVFP4 layers requantize K (E2M1 g64) and V (ISO3 g64) and rANS-encode into entropy slots (320 B header + 32 x 256 B streams + 1024 B scale tail); INT8 layers keep the raw nibble slots from PR2. Per-layer dtype dispatch replaces the all-I8 gate, so the cold pool finally compresses the layers that need it most.
- One 9536 B slot buffer serves both codecs.

**Cold-pool fixes found while validating**:
- `cold_i8` pack/restore hard-coded the 9232 B slot stride, corrupting heads past slot 0 once the buffer grew; both kernels now take `slot_bytes`.
- `can_cold_transfer` required `writer_references == 0`, which the paged store never clears for live pages, so cold transfer never ran; the gate now only requires a device replica and no pins/fork ties.
- Restore inventory treats cold pages as resident (device_kv_prefix_pages / shared variants), fixing cross-sequence prefix reuse.

**Upstream sync**: `fix(runtime): bound shared capture target expansion` cherry-picked.

### Measured results (RTX 5090D 32 GB + 8 GB host, YaRN + MTP, --cold-policy window)
| Config | Context ceiling | Device KV runtime |
|---|---|---|
| all-INT8 (pre-PR) | 272k | 10.76 GiB |
| **10L NVFP4 + cold pool** | **362k+ (370k KV capacity)** | 10.73 GiB |

Needle recall verified at every point; ceiling is the device memory wall (380k needs 11.79 GiB > 11.60 GiB available).

### Known limitations
- **Host cold-slot payloads** (`--cold-policy host`) were implemented and functionally verified (needle recall with 48 pages compressed to pinned host memory) but **reverted**: the device page pool stays at the logical capacity, so host slots do not reduce device memory. Shrinking the pool requires prefill to allocate pages chunk-progressively (currently it plans the full request up front), which is a larger scheduling change.
- **Concurrent reuse of catalogued checkpoints containing cold pages** is not supported (`cold checkpoint page has no source bookkeeping`); active-source reuse works (restore inventory fix), single-sequence operation unaffected.
- `--cold-policy host` is incompatible with speculative backends (their KV must stay resident).

### Dependency
Depends on PR1-5. Merge last.


Astrangemaninhere:
Hi @Neroued — thanks for looking at the PRs. I noticed the six PRs (#126-#131) were closed; if it is because the stacked split does not merge cleanly against master, I am happy to rebase them (or fold everything into this one) against the latest master and reopen. The NVFP4 cold-pool work in #131 is the cumulative one: per-layer KV storage, the entropy-coded cold pool, YaRN factor-4, on-demand graphs and the DFlash2 backend, plus the DFlash2 graph-prep segfault fix and --spec auto resolution. Let me know if a rebased single PR would be preferable.

PR #141, opened 2026-08-31
fix(runtime): exclude the admission source from guided pressure closure

### LINKED ISSUE #135: Host KV cache terminates the inference engine: "materialization pressure owner is duplicated" (3d9fda2, default --host-kv-mib affected)
## Description

At `3d9fda2` (current `master`), enabling the Host KV cache terminates the inference engine under
ordinary multi-session pressure. The active request fails with:

```text
materialization pressure owner is duplicated
```

The exception reaches the inference worker's fatal failure boundary. The HTTP server keeps running,
but the engine becomes unavailable and every subsequent request returns HTTP 503 until restart.

This is **not** #121 (`prepared pressure expansion exceeds the target arena`). That assertion is
fixed by `5e4bf31`, which this build contains; I verified it does not appear in my logs. This is a
different throw in the same subsystem, and it reproduces in **under a minute**.

Since `--host-kv-mib` **defaults to 8192**, a default `ninfer-serve` invocation is affected — the
flag has to be explicitly set to `0` to avoid it.

## User-visible impact

```text
[2026-08-31 17:59:48.557] [info] ninfer-serve: [req 32] error materialization pressure owner is duplicated
[2026-08-31 17:59:52.056] [info] ninfer-serve: [req 33] rejected phase=prepare protocol=openai_chat_completions stream msgs=57 media=0 tools=29 reasoning_effort=medium->unresolved status=503 code=service_unavailable message=inference engine is unavailable
```

Every later request is rejected identically. As in #121, the process stays alive and `/v1/models`
still answers, so a liveness check that does not perform inference does not detect the outage.

This first hit me in production after ~11 minutes of real two-agent coding-assistant traffic.

## Reproduction

Five concurrent conversations share one identical ~45,000-token prefix, each carrying its own
growing private history, with combined KV demand deliberately over the device pool so the pressure
planner has to run. Reproduces in **40–45 seconds**, every attempt.

```
--max-context 234496 --kv-capacity 234496 --max-concurrency 4 --prefill-chunk 1024
--kv-dtype fp8 --device-state-slots 0 --host-kv-mib 24576 --host-state-slots 8
--spec mtp --draft-tokens 3 --lm-head-draft --vision
```

Four of the five workers got HTTP 500 with the message above; the fifth then got the 503.

## Bisect

I first suspected value-aware shared-prefix scheduling, because **every request immediately before
the production crash used `reuse=shared_stable_prefix`**. That was wrong — disabling it does not
help. The variable that matters is `--host-kv-mib`:

| `--host-kv-mib` | `--host-state-slots` | `--max-shared-prefixes` | result |
|---|---|---|---|
| 24576 | 8 | 4 | **crash in 45 s** |
| 24576 | 8 | 0 | **crash in 40 s** |
| 0 | 0 | 0 | clean: 333 s, 40 requests |
| 0 | 0 | 4 | clean: 470 s, 52 requests |
| 0 | 0 | 4 (shipped default) | clean: 664 s, 70 requests |

So the trigger is the Host KV demotion path, and shared-prefix scheduling is not involved.

## Possible cause

Offered tentatively — I did not instrument a build, this is from reading the source.

The throw is in `src/targets/qwen3_6/impl/runtime/program_impl.h` (~line 3556), which rejects a
repeated continuation **index** (or one equal to the materialization source):

```cpp
if ((details.has_source && index == details.source_index &&
     generation == details.source_generation) ||
    std::find(details.pressure_indices.begin(), details.pressure_indices.end(), index) !=
        details.pressure_indices.end()) {
    throw std::invalid_argument("materialization pressure owner is duplicated");
}
```

The planning session in `pressure_planner.h` (~line 255) de-duplicates its owner list by
**ordinal**, not by continuation index:

```cpp
for (std::size_t index = 1; index < owners.size(); ++index) {
    if (owners[index - 1].ordinal == owners[index].ordinal) {
        throw std::logic_error("pressure planning owner ordinal is duplicated");
    }
}
```

If two distinct ordinals can reach the same continuation index once Host KV demotion
(`DemoteToHost` / `DropHostDuplicate`) puts a continuation into the candidate set more than once,
the ordinal check passes and the index check then fires. That would explain why the fault needs a
non-zero `--host-kv-mib` and is independent of shared prefixes.

## Environment

- GPU: NVIDIA GeForce RTX 5090, 32,607 MiB (`sm_120a`), driver `610.74`
- CPU: AMD Ryzen 7 9800X3D, 70 GiB RAM available to the guest
- OS: Debian GNU/Linux 13 (trixie) under **WSL2**, kernel `6.6.87.2-microsoft-standard-WSL2`
- CUDA compile/runtime: 13.3 (V13.3.73); host compiler g++ 14.2.0 (Debian 14.2.0-19)
- ninfer: `3d9fda2` (`master`), Release, `CMAKE_CUDA_ARCHITECTURES=120a`
- Artifact: `Qwen3.8-27B-nvfp4-NInfer` (`qwen3_8_27b_nvfp4.ninfer`, v2)
- Startup ledger for the crashing config:
  `resolved=234496 tokens pages=3664/14656 runtime=9.41 GiB free-after-weights=10.20 GiB free-after-startup=1.09 GiB slack=800.31 MiB device-state=0-cache+4-active host-state=8 host-kv=24.00 GiB private=8 shared=4`

The GPU drives no display (output is on the iGPU), so the full 32 GiB is available to ninfer.

## Checks performed / not verified

Performed:
- reproduced 3x on the crashing configuration; every attempt faulted within 40-45 s;
- ran the four bisect arms above, each to completion, plus an 11-minute / 70-request soak on the
  workaround config with zero errors;
- confirmed `prepared pressure expansion exceeds the target arena` (#121) does **not** appear in
  any of my logs, so this is not a recurrence of that fixed defect;
- confirmed the artifact is v2 and that `master` had no newer commit at the time of writing;
- confirmed the engine, not the HTTP layer, is what dies: `/v1/models` keeps answering after the fault.

Not verified:
- whether a *small* non-zero `--host-kv-mib` also triggers it — only `0` and `24576` were tested;
- whether `--vision`, `--kv-dtype fp8`, or `--device-state-slots 0` are necessary; my crashing and
  working configurations differ only in the Host KV flags, but I did not sweep the others;
- the proposed cause. I did not instrument a build or attach a debugger; the ordinal-vs-index
  reading is from source inspection only and may be wrong.

## Workaround

`--host-kv-mib 0 --host-state-slots 0`. Costs the re-prefill saving on evicted sessions; device-side
prefix reuse is unaffected. Happy to re-test a fix branch — the reproduction is reliable and takes
under a minute.

dnnspaul:
## Linked issue and scope

Fixes #135 — the pressure planner terminates the inference engine with
`materialization pressure owner is duplicated`; the HTTP server stays up and every subsequent
request returns 503 `inference engine is unavailable` until a manual restart.

I am filing this ahead of maintainer scoping on the issue because the defect is engine-fatal for
any deployment that reaches sustained KV pressure (three independent reports on #135, including a
`--max-concurrency 1` reproduction), and the fix restores an existing invariant rather than making
a design decision. Happy to rescope or rework per review.

## Root cause

`populate_options` deliberately gives the admission's reuse source no pressure options — the
`selected_private_source` / `selected_shared_source` checks skip it, leaving its
`eviction_choices[owner] == 0`. Every other owner walk in the planning session honors that marker:

- `prepare_expansion`: `if (eviction_choice == 0) { continue; }`
- the guidance loop (`assess_target` relief accounting): `if (eviction_choice == 0 || ...) { continue; }`
- `root_maximal_target`: copies `eviction_choices`, so the source contributes choice 0
- `seal` / `seal_capture` / target assessment: skip `choice == 0`

`guided_closure_target` (new in 3d9fda2) is the one walk that does not. Its selection loop iterates
`owner_order`, which is built from **all** owners, and calls
`inspect_pressure_successors(continuation_states[...], residual, &*protection, current)` directly.
`protection` shields only the specific state image and pages the admission claims
(`inspect_pressure_option` uses it nowhere else), so the program returns non-evicting successors
(checkpoint demotions/drops) for the source continuation. The closure then pushes them into the
source's deliberately empty option list, selects a nonzero choice for it, and
`compose_materialization` rejects the plan through the engine-fatal duplicate-owner contract check.

This was established empirically, not by reading alone: instrumenting the throw site on the
production deployment from #135 captured 8 collisions, and every one was the source-equality branch
with a non-evicting decision:

```text
[dup-owner] position=0 index=3 gen=1 has_source=1 source_index=3 source_gen=1 accumulated=[] owners=2 evicts=0
[dup-owner] position=0 index=0 gen=2 has_source=1 source_index=0 source_gen=2 accumulated=[] owners=3 evicts=0
[dup-owner] position=1 index=2 gen=2 has_source=1 source_index=2 source_gen=2 accumulated=[1:2] owners=4 evicts=0
...
```

It also explains the reports on #135: concurrency, KV dtype, vision, Host KV size and shared
prefixes are all irrelevant to the defect — they only change how often the pressure planner runs
during an admission that reuses a retained continuation, which is the actual trigger (consistent
with the `--max-concurrency 1` + `reuse=private_response_replay` reproduction reported there).

## Design

One guard at the top of the guided closure's selection loop, identical in meaning to the guard
every other owner walk already uses:

```cpp
if (options.eviction_choices[owner_index] == 0) { continue; }
```

`eviction_choices == 0` marks exactly the admission's source: `populate_options` throws
`"... has no maximal outcome"` for any other owner it cannot evict. The closure can still use every
other owner; when the residual cannot be closed without touching the source, it returns
`std::nullopt` and the caller falls back exactly as before this walk existed. No public contract,
ownership boundary, or schema changes.

Two observations left for the maintainer, deliberately **not** part of this PR:

- `compose_materialization` treats a planner-proposed duplicate as engine-fatal although both call
  sites already handle `std::nullopt`; a soft rejection would keep a future planner slip from
  taking the engine down. Separate decision, separate PR if wanted.
- The demotions the closure proposed for the source are not absurd in isolation (demote an unused
  checkpoint of the reused continuation). If that is ever intended, composition would need distinct
  source handling; this guard restores today's invariant without deciding that question.

## Verification

On the PR branch exactly as submitted (RTX 5090 32 GiB `sm_120a`, driver 610.74, CUDA 13.3
compile/runtime, g++ 14.2, Debian 13 under WSL2, `-DCMAKE_CUDA_ARCHITECTURES=120a`, Release):

```
ninja -C build ninfer_resource_manager_test ninfer_admission_policy_test
./build/tests/ninfer_resource_manager_test   # ok, exit 0
./build/tests/ninfer_admission_policy_test   # ok, exit 0
```

A/B against the fault, on an otherwise identical build with the throw site instrumented (log +
soft-reject) so collisions are countable without killing the engine, Qwen3.8-27B NVFP4,
`--max-concurrency 4 --kv-dtype fp8 --max-context 234496 --kv-capacity 234496 --host-kv-mib 24576
--host-state-slots 8 --spec mtp --draft-tokens 3 --lm-head-draft --vision`:

- 5 concurrent sessions sharing a ~45k-token prefix (the config that previously faulted in
  40–45 s): **without the guard 8 collisions in ~4 min; with it 0 collisions**, 81 requests, 0 errors.
- 3 concurrent ~130k-token appending conversations against the 234k pool (the workload that killed
  the production deployment twice): 20 min, 0 collisions, 0 errors.
- Production soak with the guard: the deployment from #135 has been serving two live coding agents
  — 324 requests, prompts up to 152,441 tokens, sustained pressure-planner activity — with zero
  collisions and zero engine failures, where the unfixed build died twice within ~11–31 minutes of
  the same traffic.

## Checks not run

- No performance claim is made and none was measured; the guard only skips an owner that could
  never contribute a composable plan.
- The multi-hour mixed soak ran on the instrumented sibling build, not the exact PR binary; the
  delta between them is diagnostics at the (now unreachable) throw site.
- Vision/CUDA execution routes are untouched by this change and were exercised only incidentally
  (the production deployment runs `--vision`).

### DIFF
--- a/src/targets/qwen3_6/impl/runtime/pressure_planner.h
+++ b/src/targets/qwen3_6/impl/runtime/pressure_planner.h
@@ -659,6 +659,13 @@ PressurePlanningSessionImpl<NINFER_QWEN36_VARIANT>::guided_closure_target(
         std::optional<Selection> selected;
         for (int destructive = 0; destructive < 2 && !selected; ++destructive) {
             for (const std::size_t owner_index : owner_order) {
+                // populate_options gives the admission's source owner no options
+                // (eviction_choices == 0), and every other owner walk skips it on that
+                // basis. Skip it here as well: proposing pressure decisions for the
+                // continuation the admission itself reuses is rejected by
+                // compose_materialization through the engine-fatal "materialization
+                // pressure owner is duplicated" contract check (#135).
+                if (options.eviction_choices[owner_index] == 0) { continue; }
                 const std::uint16_t current_choice = target.owner_choices[owner_index];
                 const PressureDecision* current =
                     current_choice == 0 ? nullptr


dnnspaul:
Superseded by `da49c0d`, which enforces the same invariant structurally (OwnerParticipation) and adds the source-pressure-protection regression scenario. Verified on the reporting deployment — see #135.

Issue #144, opened 2026-08-31
Concurrent streams re-trigger root re-prefill of parked tail contexts


wSzki:
Related: #120, #98

> **TL;DR:** The #120 fix is confirmed for its original workload, but under
> concurrent streamed load a rotation over a second cohort of stored contexts
> re-prefills the LAST context of every round from `root` (~11 s, 0 cache
> hits) once Host State descriptors saturate (24/24 used, Host KV only half
> full), while the other five restore in under 0.35 s. Replicated; depends on
> process history, not the streams alone. (The sticky duplicate-owner 500 seen
> in the same campaign is #135, fix proposed in #141 - not restated here.)
>
> All resumes kept correct markers - retention/latency/availability only, no
> output corruption. Full matrix, repro, and diagnostics below. Can attach
> logs/drivers or run bisect variants.

## Context

The `3d9fda2` fix for #120 passes its original sequential gate (19/19 resumes
restored through `private_turn_closure`, including the former round-2 collapse
point). Testing on top of that gate - concurrent streamed load, and processes
holding stored contexts created by more than one client cohort - shows one
further behavior in the materialization planner: a cache-retention and
latency finding. All 725+ resumes across every run returned both expected
markers, so output correctness is not in question. (The same campaign also
hit #135's sticky `materialization pressure owner is duplicated` 500; that
defect is tracked there with a proposed fix in #141.)

### Behavior 1 - tail re-prefill, replicated

With two concurrent streamed generations running, a rotation over a second
cohort of six stored contexts re-prefills the last context of every round from
`root`: zero cache hits, 55,073 computed tokens, 11.28-11.33 s TTFT, while the
other contexts in the same rounds restore through `private_turn_closure` at
0.07-0.33 s. The behavior did not appear on a fresh process (18/18 restored),
nor after a different warm-up history, so it depends on process history, not
on the streams alone. In the replicated run the Host State descriptor budget
was full (24 of 24 slots) with 23.2 of 48 GiB Host KV in use - the bytes are
half free while the descriptors are exhausted, and the telemetry does not say
which constraint the planner acted on. Unlike `d9dbe1c`,
`selected_maximal_fallback` is false on every collapsed resume and the parked
estate is left standing.

Selected fields for a collapsed resume (schema 18, identical shape in the
47-minute soak and the replication):

```json
{
  "result": {
    "prompt_tokens": 55073,
    "prefix_cache_hit_tokens": 0,
    "computed_prefill_tokens": 55073,
[...]

PR #146, opened 2026-08-31
fix(runtime): probe Host KV allocator geometry in guided pressure closure

Astrangemaninhere:
## Linked issue and scope

Addresses #144: under concurrent streamed load a rotation over a second cohort of stored contexts re-prefills the last context of every round from `root` (~11 s TTFT, 0 cache hits) once the Host State descriptors saturate, even though Host KV is only half full and evicting one parked context would unblock the request.

## Root cause

`blocked_host_allocation_bytes` is only ever assigned inside `compose_materialization`, and that function requires its input candidate to be unblocked, so every candidate the planner sees carries zero. The previous byte fold in the guided closure and guidance therefore did nothing: **allocator geometry is not representable in the guided residual at all**. When the `DemoteToHost` requests of a pressure target cannot be allocated even though free Host KV bytes are ample, the closure closed immediately on an infeasible target, the exact assessment rejected it, and the bounded search fell back to a root re-prefill. The observed `stop_reason: time_budget`, `targets_evaluated: 12`, `selected_maximal_fallback: false` and a standing parked estate match this path exactly.

## Change

- New `ProgramImplCore::guided_host_allocation_feasible()`: a host-side projection of the candidate plus the selected pressure decisions into allocation requests (`DemoteToHost`) and releases (`DropHostDuplicate` replicas and eviction last-reference pages), answered by the real extent allocator (`can_allocate_after_page_releases`) — not by subtracting bytes.
- The guided closure's per-step residual and the target guidance residual keep a non-zero component while that probe fails, so the closure can drive toward relief (e.g. evicting one whole parked context) instead of closing on a target the allocator rejects; the exact composition still re-validates before any plan is sealed.

## Test

- Resource-manager harness mirrors the probe with release bytes; regression test `guided closure relieves allocator geometry block` asserts a single-owner eviction relieves an allocator-geometry block. Verified: the test fails on the pre-fix closure behavior and passes with the fix; the full `test_resource_manager` suite passes.
- Built in WSL with CUDA 13.3 and run on an RTX 5090 against `qwen3_8_27b_nvfp4` (real weights): `ninfer_qwen3_6_context_store_test`, `prefix_real` scenarios `pressure-resume` and `private-checkpoint-pressure` all pass; the Qwen3.6/3.8 session template instantiations compile cleanly.

### DIFF
--- a/src/targets/qwen3_6/impl/runtime/pressure_planner.h
+++ b/src/targets/qwen3_6/impl/runtime/pressure_planner.h
@@ -533,6 +533,14 @@ PressurePlanningSessionImpl<NINFER_QWEN36_VARIANT>::guided_closure_target(
                                         std::optional<std::size_t> override_owner,
                                         const PressureDecision* override_decision) {
         detail::PhysicalDelta pressure;
+        std::vector<const ContinuationHandle*> selected_private_handles;
+        std::vector<const PressureDecision*> selected_private_decisions;
+        std::vector<const SharedPrefixHandle*> selected_shared_handles;
+        std::vector<const PressureDecision*> selected_shared_decisions;
+        selected_private_handles.reserve(owners.size());
+        selected_private_decisions.reserve(owners.size());
+        selected_shared_handles.reserve(owners.size());
+        selected_shared_decisions.reserve(owners.size());
         for (std::size_t index = 0; index < owners.size(); ++index) {
             const PressureDecision* decision = nullptr;
             if (override_owner && *override_owner == index) {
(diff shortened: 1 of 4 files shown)


Astrangemaninhere:
You are right, and investigating this further exposed that the byte-relief fold is not merely approximate — it is dead code, which is worse.

`blocked_host_allocation_bytes` is only ever assigned inside `compose_materialization` (`program_impl.h:3781`), and that function **requires its input candidate to have `blocked_host_allocation_bytes == 0`** (it throws otherwise). Every candidate the planner and the guided closure see is the pre-composition candidate, so the fold in `guided_closure_target()`/`guidance()` reads a constant zero. The allocator-geometry constraint is therefore the one resource that is *not representable in the guided residual at all*: bytes and slots all fit, the closure closes immediately with the empty pressure target, the exact assessment (post-composition `blocked != 0`) rejects it, and the bounded search falls back to a root re-prefill. That is exactly the `time_budget` / 12-target / non-maximal telemetry in #144.

Fix direction (agreeing with your recommendation): add a Program-side geometry probe that projects the candidate plus the closure's currently selected decisions into `HostKVAllocationRequest`s and `HostKVPageReplicaRelease`s (demote requests, drop-host-duplicate releases, and eviction last-reference releases) and answers `can_allocate_after_page_releases`. The closure residual then carries a non-zero “geometry still blocked” component until the probe passes, so the closure can actually drive toward “evict one parked context” instead of giving up, while the exact composition still re-validates before any plan is sealed. I will replace the byte subtraction with this probe in both the closure residual and `guidance()`, mirror it in the resource-manager test harness (geometry probe instead of byte relief), and update the PR.

Thanks for the review — the byte-subtraction version would have shipped a no-op with a misleading name.

Astrangemaninhere:
Implemented as recommended. New commit `93bd6d6` adds `ProgramImplCore::guided_host_allocation_feasible()`, which projects the candidate plus the selected decisions into real `HostKVAllocationRequest`s and `HostKVPageReplicaRelease`s (DemoteToHost requests, DropHostDuplicate replica releases, eviction last-reference pages) and answers with `can_allocate_after_page_releases` — no byte arithmetic. The closure residual and `guidance()` keep a non-zero component while that probe fails, and the exact composition still re-validates before any seal.

This also exposed that the previous byte fold was worse than approximate: `blocked_host_allocation_bytes` is only assigned inside `compose_materialization`, which requires an unblocked input, so every planner-visible candidate carried zero and the fold was a no-op — geometry was simply absent from the guided residual. The probe fixes that gap.

Verified on this machine (WSL, CUDA 13.3, RTX 5090, real `qwen3_8_27b_nvfp4` weights): `test_resource_manager` (including the new geometry-block regression test), `ninfer_qwen3_6_context_store_test`, and both `prefix_real` GPU scenarios (`pressure-resume`, `private-checkpoint-pressure`) pass.

Issue #143, opened 2026-08-31
Future: NVMe cold tier under Device/Host planner for parked continuations


steve8697:
Related: #142 (system-frontier write candidate). This is a **later** capacity idea, not a substitute for that serve-layer fix.

## What we are asking

Would you consider, after the current Device + pinned-Host catalog is settled, a **cold NVMe tier** under the same planner — not a second cache product.

Today the degradation graph is already monotonic:

Device replica → pinned Host State/KV → drop checkpoint → evict owner

A natural extra step is:

Device (hot) → pinned DDR5 Host (warm, `--host-kv-mib`) → NVMe (cold) → evict

vLLM/LMCache, TensorRT-LLM host offload, and NVIDIA ICMSP all use this shape for **parked** context. NInfer is stricter (exact frontier + full StateImage + GDN recurrent bundle), so a disk tier only makes sense if it stores the **same checkpoint object** the Host extent store already knows, not raw KV pages alone.

## What it would help, and what it would not

Measured on this box (RTX 5090 32 GiB, WSL2 64 GiB DDR5, C=2, INT8, Qwen3.8-27B NVFP4):

| already true today | SSD/DDR5 would **not** change |
|---|---|
| Live decode stays on Device | Tokens/s of an active lane |
| `--max-concurrency 2` is the GPU lane cap | Two 150k chats decoding at once (KV pool is still one 245k Device pool) |
| Same-turn `private_turn_closure` is already ~90–125 ms | First-token of a hot continuation |

| gap | why a cold tier might help |
|---|---|
| Returning to a **parked** 80–120k agent chat after other sessions filled Device+Host | Avoid 20–50 s `root` re-prefill. Fork `#73` already showed 118k Device miss → host restore **422 ms vs 53 s**. NVMe would be slower than pinned DDR, but still far cheaper than prefill if sequential restore is ~1–3 s for a few GiB. |
| More retained **session count** at C=2 | C=2 only has 2 active + a handful of Device/Host checkpoints. Extra cold slots raise the chance the next Hermes/OpenCode window is a restore instead of `root`. That is reuse probability, not live batch width. |
| Shared system prefix once #142 exists | One published 8–30k `shared_stable_prefix` could sit in Host or NVMe and seed many siblings without occupying Device KV. |

So: **more parked conversations survive**, not **higher simultaneous decode concurrency**. Worth saying explicitly so this is not read as “SSD makes C=8 on 32 GiB”.

## Constraints that match current contracts

[...]

PR #149, opened 2026-09-01
feat(runtime): NVMe cold tier (ColdPolicy::Disk) under the cold-pool planner

Astrangemaninhere:
## NVMe cold tier under the cold-pool planner (ColdPolicy::Disk)

Adds a disk placement to the existing cold-pool degradation chain:
Device (hot) -> pinned Host (warm) -> **NVMe file (cold)** -> evict.

### What it does

- `ColdPolicy::Disk` + `--cold-policy disk` with `--cold-disk-path` and
  `--cold-disk-bytes` (default 32 GiB). Empty path uses the system temp dir
  (WSL2: keep it on ext4, not `/mnt/c`).
- Cold pages spill to **per-layer files**: each slot is a fixed-stride unit
  (compressed bytes, rANS or raw codec unchanged), so the file offset is
  `slot * stride`. The device slot pool becomes a working set; capacity is
  bounded by the disk budget instead of `max_cold_pages`.
- Egress: after the pack kernels commit a slot, its bytes are mirrored to the
  file (D2H via the device stream, then write). Restore reads the file back
  before the decode kernels run.
- **Async prefetch**: `warm_cold_prefix` prefetches every pending slot on the
  transfer stream with double-buffered pinned staging, so the H2D legs of the
  warm path overlap with the previous decode step; the decode stream waits
  once before restoring the batch.
- `restore_cold_page` gains a `disk_prefetched` flag; the checkpoint/rewrite
  path falls back to the synchronous read.

### Why

Responds to the design in upstream #143 (NVMe cold tier under the
Device/Host planner). Same shape as vLLM/LMCache, TRT-LLM host offload and
NVIDIA ICMSP: more **parked** conversations survive (reuse probability),
not higher live decode concurrency. Restore of a few GiB from NVMe is far
cheaper than re-prefilling an 80-120k agent chat (upstream #73 measured
host restore at 422 ms vs 53 s prefill).

### Notes

- Reads stay exact-identity; the slot codecs are byte-identical to the
  device pool (no approximate / block-hash cache).
- Live decode never streams from the file; cold restore is admission-time
  materialize-to-device, then run.
- Opt-in, default off: current two-tier behavior is unchanged.

Verified: builds clean, baseline ppl unchanged (1.3353 on the zh corpus),
disk spill/restore exercised with `--cold-policy disk --cold-disk-path /tmp`.

## Verification status

- Functional (RTX 5090 / WSL2, `--cold-policy disk --cold-disk-path /tmp`):
  egress compressed 17 prefix pages and mirrored them (1,296,896 B/layer,
  exact `pages * slot_stride`); restore + prefix reuse verified (4.3k-token
  prompt 2.5 s first request, 0.9 s on reuse).
- Build: compiled and linked in the integration tree this branch is based on
  (engine + ops). **Limitation**: the rebased branch (on latest master) has
  not been built standalone yet — the same 8-file diff was built and linked
  in the pre-rebase tree; a clean-master build is pending.
- Codex review findings (6) addressed in the latest commit; the DFlash2 two
  are pre-existing integration-tree fixes included in the diff.
- This implements the proposal in #143; direction confirmed with the
  maintainer is still pending (comment posted on #143).


Astrangemaninhere:
All six review findings are addressed in the updated commit (1f40594). Functional verification on the 5090/WSL2 box with `--cold-policy disk`:

- egress: 17 prefix pages compressed and mirrored (1,296,896 B/layer = 17 * slot stride, exact), device slot released after the mirror
- restore: cold pages read back and decoded, prefix reuse across requests works (long 4.3k prompt: 2.5 s first, 0.9 s on reuse)
- file mode `w+b`, staging element addressing, and the two Window-only gates (`enqueue_cold_compressions`, warm-on-retain) now include the Disk policy

Per finding:

1. **Run cold maintenance for Disk** — both gate sites (enqueue trigger + retained-source warm) now accept `ColdPolicy::Disk`; disk compression is exercised in the run above.
2. **Staging buffer addressing** — the egress D2H and the restore H2D now use `cold_disk_staging[0]` (the pinned allocation), not the array name.
3. **Open spill files read/write** — files open with `"w+b"`; restore reads succeed (verified by the reuse run).
4. **Size the disk tier** — rather than inflating the device slot pool, the disk tier now uses an independent monotonic **file-slot counter**: egress mirrors at `file_slot * stride` and releases the device slot (the file is the durable copy); restore allocates a temporary device staging slot, reads the file back, decodes, and releases it. Capacity is bounded by the disk budget; the device pool stays the small working set.
5. **DFlash2 draft width** — explicit `--spec dflash2` without `--draft-tokens` now normalizes to the fixed 7-draft block in the planner (validation accepts 0/7; the planning inputs map 0 -> 7), so the sequence planner never sees a zero draft window.
6. **DFlash2 long-context demotion** — the submission loop now applies the same `kSpecDemoteTokens` predicate as the budget loop (extent 0 past the threshold), keeping the submitted extent within the computed target envelope.

Thanks for the review.

PR #152, opened 2026-09-01
feat(serve): automatic shared-prefix write at the system/developer frontier

### LINKED ISSUE #142: Agent sibling sessions miss shared prefix without prompt_cache_breakpoint
## Summary

Agent clients (Hermes, OpenCode, Pi, etc.) almost never send `prompt_cache_breakpoint`. On current master, that means **sibling conversations that share a long system/tools head always miss** and re-prefill the whole head.

The official catalog already has the right hit path (`shared_stable_prefix`). The gap is the **default OpenAI write candidate**, which lands on the last content part (usually the unique user turn), not on the stable system block.

This is **not** a request to merge #73 / #90 as a second store. It is a request to place one automatic write candidate at the rendered system-block frontier, so the existing planner can publish a shared prefix that those clients actually share.

## Measured on RTX 5090, Qwen3.8-27B NVFP4, master `3d9fda2`

Serve flags (exclusive one-GPU lane): `--max-concurrency 2 --kv-dtype int8 --vision --host-kv-mib 16384 --host-state-slots 8 --device-state-slots 2 --max-shared-prefixes 4 --max-private-continuations 4 --spec mtp --draft-tokens 3 --lm-head-draft`, ctx/kv 245760.

Chat Completions, `reasoning_effort=low`, `max_tokens=8`. Authoritative numbers from the serve `done` line.

| scenario | `reuse=` | cache/prompt | TTFT |
|---|---|---|---|
| same conversation, next turn (~2k) | `private_turn_closure` | 2188/2209 | **92 ms** (cold 559 ms) |
| 16k shared system, **no** breakpoint, two sibling users | both `root` | 0/16484 | **2363 / 2363 ms** |
| same 16k system + `prompt_cache_breakpoint` on the system part | `shared_stable_prefix` | 16466/16485 | **91 ms** |
| C=2 concurrent siblings after that prefix was published | both `shared_stable_prefix` | 16466 | 89 / 223 ms |
| same conversation, `reasoning_effort` low → xhigh | `root` | 0 | 347 ms |
| 3×32k private then return to A | `private_turn_closure` | 32776/32797 | **125 ms** (cold 5365 ms) |

So: same-turn reuse is already excellent. Shared-prefix reuse is also excellent **once a legal write exists**. Default OpenAI implicit write does not create that write for the agent-sibling shape.

## Why the default implicit candidate misses

`docs/serving.md` (OpenAI prompt caching): omitted `prompt_cache_options` creates a default implicit candidate at the **latest representable content boundary**. In `apply_openai_prompt_cache_policy` that is the last content part of the last turn — for a normal chat request, the unique user message.

Request A is then `system(16k) + user("SIBA")` published as one prefix. Request B is `system(16k) + user("SIBB")`. Exact prefix identity diverges at the last user tokens, so B is `root` even though 16k tokens are identical.

`allow_engine_automatic_shared_prefixes` is also forced off after OpenAI policy is applied (`openai_common.cpp`), which is consistent with “OpenAI already defined the write policy” — but that policy’s implicit target is the wrong frontier for this workload.

Fork #90 (`seed_prefix` / `--prefix-cache-mib`) solved the same gap with a **separate** GPU seed store. That is a second catalog and would fight the current planner. The behavior we want is only: **capture at the system-block frontier using the existing shared-prefix machinery**.

## Suggestion (smallest change)

Keep the OpenAI explicit-breakpoint contract. Add an extra default write candidate when Chat Completions / Responses has a leading system (or tools+system) block and the client did **not** already mark it:

1. After rendering, take the token frontier just past the leading system / tools block (the same frontier #90 used).
2. If that frontier is above a floor (256 tokens, or whatever the value model already uses), emit a `SharedStablePrefix` candidate with `DefaultAutomatic` / `RequestedAutomatic` evidence.
3. Do **not** disable the existing last-content implicit candidate; a same-turn continuation still wants a later frontier.
4. Reads of an already-published prefix should keep working without the client repeating a marker (this already matches serving.md).

That is one extra legal write (still ≤ 4). It does not add `PrefixSeedStore`, `--prefix-cache-mib`, `--kv-host-cache-mib`, or overlay.

Optional serve flag if you want it opt-in: `--auto-system-shared-prefix` default on for agent serving, off for strict OpenAI-implicit-only.

## What we are not asking for

- Do not resurrect #73’s separate content-addressed host KV store next to `--host-kv-mib`. Same-axis, two eviction policies.
- Do not resurrect #90’s device arena as a third sequence-state home.
- Overlay vision residency is a different VRAM trade; not required for this.

Private continuation + host-kv already covers “same conversation came back.” The remaining hole at C=2 is **two chats, one long head, no client marker**.

Happy to re-run the sibling probe on a patch.

Astrangemaninhere:
Closes #142

## What

Agent clients (Hermes, OpenCode, Pi, etc.) rarely send `prompt_cache_breakpoint`; on master they always miss the shared system/tools head. This adds **one automatic write candidate at the leading system/developer frontier** so the existing shared-prefix machinery can publish a prefix those clients actually share.

- Keeps the OpenAI explicit-breakpoint contract and the existing last-content implicit candidate (both writes coexist).
- Opt-out flag `--no-auto-system-shared-prefix` (extra write on by default) for the Responses path; the Chat path currently uses the default-on policy value (consistent with the issue discussion; wiring the flag into the Chat parse signature was deliberately left out to keep the diff minimal).
- No second store, no new catalog, no engine changes - the planner's existing `shared_stable_prefix` path does the work.

## Per-file rationale (8 files, +65/-7, serve layer only)

1. **`src/serve/openai_common.h`** (+5) - `OpenAIPromptCachePolicy` gains `auto_system_shared_prefix` (default `true`). The policy is the single switch that reaches both Chat and Responses without threading a new parameter through every parser.
2. **`src/serve/openai_common.cpp`** (+31/-1) - the actual feature:
   - Locates the **end of the contiguous leading system/developer run** (not the first turn): requests that share both a system and a developer turn reuse the whole instruction prefix.
   - Publishes the `DefaultAutomatic` candidate on that last leading turn when unmarked (the existing explicit-breakpoint contract is untouched).
   - **Reserves one of the four frontend marker slots** for the new candidate (`frontend.cpp` rejects >4 markers); the explicit-write selection shrinks by one when the candidate is distinct, so a request with 3 explicit breakpoints + both automatic writes stays at 4.
3. **`src/serve/serve_options.h`** (+3) - `ServeOptions` gains `auto_system_shared_prefix = true`.
4. **`src/serve/serve_options.cpp`** (+4/-1) - parses `--no-auto-system-shared-prefix` and documents it in the usage text.
5. **`src/serve/openai_responses.h`** (+8/-1) - `OpenAIResponsesPromptRequest` carries the parsed `cache_policy` so the policy can be applied **after** prompt resolution (at parse time the leading turns are still in `input_turns`/`instructions`, not in `generation.messages`).
6. **`src/serve/openai_responses_request.cpp`** (+11/-3) - stores the policy on the prompt instead of applying it to the empty message list; the parser signature gains the opt-out bool from the HTTP layer.
7. **`src/serve/openai_responses_state.cpp`** (+7) - `resolve_openai_responses_prompt` applies the policy once `generation.messages` is assembled (instructions + input turns), which is the earliest point the leading-instruction candidate can be placed correctly. (Fixes the review finding that Responses requests saw no system/developer turns.)
8. **`src/serve/openai_responses_http.cpp`** (+3/-1) - passes `options_.auto_system_shared_prefix` into the parser; without this one line the flag would exist but never reach Responses requests.

## Verification

- Build: clean on master `da49c0d` (WSL2, CUDA 13.3, sm_120a).
- Serve smoke: sibling Chat requests sharing a long system head reuse the published prefix without any client breakpoint; Responses requests place the candidate after the resolved leading instructions.
- `--no-auto-system-shared-prefix` restores strict OpenAI-implicit behavior on the Responses path.

## Follow-ups agreed with the issue author (#142 discussion, not blockers)

- **Frontier**: if the Qwen template renders tools before the system block, the candidate should sit after the combined tools+system prefix. Open to folding in.
- **Floor**: skip the extra write below ~256 tokens of shared head.
- @steve8697 offered to re-run the 16k sibling probe against this patch.

### DIFF
--- a/src/serve/openai_common.cpp
+++ b/src/serve/openai_common.cpp
@@ -144,8 +144,25 @@ void apply_openai_prompt_cache_policy(GenerationRequest& request, OpenAIPromptCa
     const bool automatic_enabled =
         policy.automatic != OpenAIPromptCacheAutomatic::Disabled && automatic_target != nullptr;
     const bool automatic_merges_explicit = automatic_enabled && automatic_target->has_value();
-    const std::size_t explicit_write_slots =
+    // Issue #142: the leading-instruction automatic candidate is a distinct
+    // marker when the leading run is unmarked, so it consumes one of the four
+    // frontend marker slots (frontend.cpp rejects more than four).
+    std::size_t leading_instruction = static_cast<std::size_t>(-1);
+    if (policy.auto_system_shared_prefix) {
+        for (std::size_t index = 0; index < request.messages.size(); ++index) {
+            const ChatRole role = request.messages[index].role;
+            if (role != ChatRole::System && role != ChatRole::Developer) { break; }
+            leading_instruction = index;
+        }
+    }
+    const bool system_candidate =
+        policy.auto_system_shared_prefix && automatic_enabled &&
+        leading_instruction != static_cast<std::size_t>(-1) &&
+        !request.messages[leading_instruction].cache_boundary_after.has_value();
+    const std::size_t base_slots =
         !automatic_enabled || automatic_merges_explicit ? 4U : 3U;
+    const std::size_t explicit_write_slots =
+        base_slots - (system_candidate ? 1U : 0U);
     const std::size_t first_selected = explicit_boundaries.size() > explicit_write_slots
                                            ? explicit_boundaries.size() - explicit_write_slots
                                            : 0U;
@@ -169,6 +186,18 @@ void apply_openai_prompt_cache_policy(GenerationRequest& request, OpenAIPromptCa
             automatic_target->value().evidence |= evidence;
         } else {
             *automatic_target = CacheBoundary{.evidence = evidence};
+        }
+        // Issue #142: a second automatic candidate at the leading
+        // system/developer frontier (agent siblings share the long head).
+        // Mark the end of the contiguous leading instruction run so requests
+        // sharing both system and developer turns reuse the whole prefix.
+        if (policy.auto_system_shared_prefix &&
+            leading_instruction != static_cast<std::size_t>(-1)) {
+            ChatTurn& turn = request.messages[leading_instruction];
+            if (!turn.cache_boundary_after) {
+                turn.cache_boundary_after = CacheBoundary{
+                    .evidence = ninfer::SharedCandidateEvidence::DefaultAutomatic};
+            }
         }
     }
     // OpenAI already defines the automatic/explicit write policy for every request. Existing
--- a/src/serve/openai_common.h
+++ b/src/serve/openai_common.h
@@ -20,6 +20,11 @@ enum class OpenAIPromptCacheAutomatic : std::uint8_t {
 
 struct OpenAIPromptCachePolicy {
     OpenAIPromptCacheAutomatic automatic = OpenAIPromptCacheAutomatic::Default;
+    // Issue #142: additionally publish a shared-prefix candidate at the
+    // leading system/developer frontier so agent sibling sessions share a
+    // long head even without a client marker. The last-content implicit
+    // candidate stays untouched.
+    bool auto_system_shared_prefix = true;
 };
 
 [[nodiscard]] bool parse_openai_prompt_cache_breakpoint(const nlohmann::json& value,
--- a/src/serve/openai_responses.h
+++ b/src/serve/openai_responses.h
@@ -34,6 +34,10 @@ struct OpenAIResponsesPromptRequest {
     std::vector<nlohmann::json> input_items;
     std::optional<std::string> instructions;
     std::optional<std::string> previous_response_id;
+    // Applied after prompt resolution assembles generation.messages (the
+    // policy needs the resolved system/developer turns to place the
+    // leading-instruction candidate).
+    std::optional<OpenAIPromptCachePolicy> cache_policy;
 };
 
 struct OpenAIResponsesCreateRequest {
@@ -72,7 +76,9 @@ struct BuiltOpenAIResponse {
 };
 
 OpenAIResponsesCreateRequest parse_openai_responses_create_request(const nlohmann::json& body,
-                                                                   const RequestLimits& limits);
+                                                                   const RequestLimits& limits,
+                                                                   bool auto_system_shared_prefix =
+                                                                       true);
 
 OpenAIResponsesPromptRequest
 parse_openai_responses_input_tokens_request(const nlohmann::json& body,
--- a/src/serve/openai_responses_http.cpp
+++ b/src/serve/openai_responses_http.cpp
@@ -245,7 +245,8 @@ void HttpServer::handle_responses(const httplib::Request& req, httplib::Response
     try {
         RequestLimits limits;
         limits.default_max_tokens = options_.default_max_tokens;
-        request = parse_openai_responses_create_request(parse_json_body(req), limits);
+        request = parse_openai_responses_create_request(
+            parse_json_body(req), limits, options_.auto_system_shared_prefix);
         validate_openai_model(request.prompt.model, public_model_id_);
         resolved = resolve_openai_responses_prompt(request.prompt, openai_responses_store_, id,
                                                    request.store);
--- a/src/serve/openai_responses_request.cpp
+++ b/src/serve/openai_responses_request.cpp
@@ -1158,16 +1158,21 @@ void validate_common_top_level(const Json& body, bool create) {
 } // namespace
 
 OpenAIResponsesCreateRequest parse_openai_responses_create_request(const Json& body,
-                                                                   const RequestLimits& limits) {
+                                                                   const RequestLimits& limits,
+                                                                   bool auto_system_shared_prefix) {
     require_object(body);
     validate_common_top_level(body, true);
     reject_unsupported_platform_fields(body);
-    const OpenAIPromptCachePolicy cache_policy = parse_openai_prompt_cache_policy(body);
+    OpenAIPromptCachePolicy cache_policy = parse_openai_prompt_cache_policy(body);
+    cache_policy.auto_system_shared_prefix = auto_system_shared_prefix;
 
     ParsedPromptFields parsed = parse_prompt_fields(body, limits);
-    apply_openai_prompt_cache_policy(parsed.prompt.generation, cache_policy);
     OpenAIResponsesCreateRequest out;
     out.prompt              = std::move(parsed.prompt);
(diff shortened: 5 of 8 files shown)


Astrangemaninhere:
这次的pr修改程度较小，希望能得到采纳；如果不能接受，希望能给出意见。

steve8697:
Thanks for landing this as a small serve-only patch, and for calling out the 16k sibling probe.

We can re-run the same Chat Completions probe we used in #142 (16k shared system, **no** `prompt_cache_breakpoint`, two sibling users; authoritative `reuse=` / `cache=` / `ttft=` from the serve log). Pass criteria on our 5090 box:

- sibling B should be `shared_stable_prefix` with cache ≈ system tokens, TTFT in the ~100 ms class (we measured 91 ms with an explicit breakpoint on master `3d9fda2`; without a marker it was `root` / 2.4 s)
- same-turn continuation should still be `private_turn_closure`
- `--no-auto-system-shared-prefix` should restore the `root` miss on the sibling

Two practical notes, not blockers:

- This PR currently shows **CONFLICTING** against current `master` (`a16b644`, ~110 commits past our last 9811 build, mostly DFlash2). We will apply the patch on a clean 9811 worktree / rebase if needed; if that fails we will say so instead of inventing numbers.
- This machine cannot dual-run `:9810` (daily fork) and a test serve. The probe waits on one GPU window.

Will post the table on this PR when the run finishes. Chat Completions is the path we can speak to; we did not exercise Responses in the original issue.

steve8697:
16k sibling probe on this patch, RTX 5090 32GB, exclusive `:9811`.

**Tree:** cherry-pick of `937abea` onto our last 9811 baseline `3d9fda2` → `4cdca00`. We did not rebase onto current `master` (`a16b644`, still CONFLICTING).

**Compile:** as submitted, `openai_responses.h` uses `OpenAIPromptCachePolicy` without including `openai_common.h`. `openai_responses_request.cpp` includes the header first, so that TU fails. We added `#include "serve/openai_common.h"` only; no other local edits.

Serve flags: `--host-kv-mib 16384 --host-state-slots 8 --device-state-slots 2 --max-shared-prefixes 4 --max-private-continuations 4 --max-concurrency 2 --kv-dtype int8 --vision --max-context 237568`, MTP3. Chat Completions is the path from #142; Responses included because that is where the CLI opt-out is wired. Authoritative `reuse=` / `cache=` / `ttft=` from the serve log.

### 1. Default ON (no client `prompt_cache_breakpoint`)

| scenario | reuse | cache | prompt | hit% | ttft ms | wall s |
|---|---|---|---|---|---|---|
| chat S1 cold 2k | `root` | 0 | 2193 | 0.0 | 320 | 0.381 |
| chat S1 warm continue | `private_turn_closure` | 2188 | 2209 | 99.0 | **86** | 0.138 |
| chat S2a sib A, implicit 16k | `root` | 0 | 16484 | 0.0 | 2402 | 2.456 |
| chat S2a sib B, implicit 16k | `root` | 0 | 16485 | 0.0 | **2428** | 2.483 |
| chat S2b sib A, explicit breakpoint | `root` | 0 | 16484 | 0.0 | 2493 | 2.530 |
| chat S2b sib B, explicit breakpoint | `shared_stable_prefix` | 16466 | 16485 | 99.9 | **108** | 0.146 |
| chat S4 conc CONC1 (after S2b) | `shared_stable_prefix` | 16468 | 16484 | 99.9 | 240 | 0.173 |
| chat S4 conc CONC2 (after S2b) | `shared_stable_prefix` | 16468 | 16484 | 99.9 | 103 | 0.295 |
| responses S2a sib A, implicit 16k | `root` | 0 | 16484 | 0.0 | 2404 | 2.459 |
| responses S2a sib B, implicit 16k | `root` | 0 | 16485 | 0.0 | 2495 | 2.550 |

Historical same box on `3d9fda2` without this patch: S2a both `root` / 2363 ms; S2b sib B `shared_stable_prefix` / 91 ms / cache=16466. Same-turn was `private_turn_closure` / 92 ms.

### 2. `--no-auto-system-shared-prefix`

| scenario | reuse | cache | prompt | ttft ms |
|---|---|---|---|---|
| opt-out chat S2a sib A/B | `root` / `root` | 0 | 16485 / 16486 | 2404 / 2428 |
| opt-out chat S1 warm continue | `private_turn_closure` | 2189 | 2210 | 110 |
| opt-out responses S2a sib A/B | `root` / `root` | 0 | 16485 / 16486 | 2426 / 2425 |

### Verdict

- **Same-turn is fine.** `private_turn_closure` still hits (~86–110 ms). Last-content implicit write is intact.
- **Explicit breakpoint still works.** S2b / S4 match the #142 control: sibling B `shared_stable_prefix`, cache ≈ system tokens, TTFT in the ~100 ms class vs ~2.4 s `root`.
- **The #142 case does not.** Chat and Responses 16k siblings with **no** client marker both stay `root` / ~2.4 s, same as unpatched master. Hermes / Pi / OpenCode would still re-prefill the shared system head.

So the extra write is not becoming a published `shared_stable_prefix` that a second user can reuse.

Likely reason, from reading current planner + this diff (not from extra instrumentation):

- The new candidate is `CacheBoundary{ .evidence = DefaultAutomatic }` on `ChatTurn::cache_boundary_after` (`MessageBoundary`).
- In `resource_manager.h`, `DefaultAutomatic` is only a surplus candidate. `pressure_capable` / `explicit_shared_credit` require `ExplicitBoundary` or `RequestedAutomatic`. The value model then has to beat baseline + split cost before anything is written into the shared catalog. That is already why today's last-content implicit write does not help siblings.
- The control that *does* publish is S2b: `prompt_cache_breakpoint` on the system **part** (`ExplicitBoundary` + `LeadingInstructionBoundary`).

`--no-auto-system-shared-prefix` could not be shown to restore a Chat miss, because Chat never hits in the first place, and the Chat parser still does not take `options_.auto_system_shared_prefix` (as the PR text said). Responses S2a also missed with the flag off.

Happy to re-run the same table if you:

1. `#include "serve/openai_common.h"` in `openai_responses.h`
2. Give the leading-instruction candidate `RequestedAutomatic` or `ExplicitBoundary` evidence so the planner treats it as declared
3. Place it as a leading-instruction **part** marker (the S2b path), not only a turn `MessageBoundary`
4. Optionally thread the CLI flag into the Chat parser

Raw: serve log `reuse=` lines from this run, ctx 237568, INT8, C=2.

Issue #154, opened 2026-09-01
Malformed UTF-8 from the sampler terminates the inference engine for the life of the process


kitaekatt:
> **Note on this report.** I have rewritten this description in place. It originally led
> with an argument that `valid_utf8_prefix_size` and `terminalize` contradict each other,
> and it understated the impact as a failed request. **The inconsistency argument was
> wrong and I have withdrawn it** (see "A correction I owe you" below), and the impact is
> considerably worse than I first described. The two comments below are the corrections as
> I found them; this description now incorporates both, so the comments are history rather
> than something you need to read first. Sorry for the churn.

## Summary

When the sampler produces a byte sequence that is not valid UTF-8, the Qwen3.6 frontend
decoder throws, and **the throw terminates the inference engine for the lifetime of the
process.** Every subsequent request from every client is rejected with
`service_unavailable`; the HTTP listener stays up and never recovers. A restart is the
only remedy.

It reproduces at your own validated evaluation configuration, with no unusual sampling
parameters, on ordinary prompts.

## Environment

- `6e8b2e2a` (master), built Release, `CMAKE_CUDA_ARCHITECTURES=120a`
- RTX 5090, WSL2
- Qwen3.8-27B NVFP4 artifact, served as `qwen3.8-27b`

## What happens

`valid_utf8_prefix_size` (`src/targets/qwen3_6/impl/frontend/frontend.cpp:402`) is correct
for a *truncated* sequence: when the declared length runs past the buffer it returns the
offset so far (`:426`), and `feed_token_bytes` (`:606-610`) keeps the remainder in
`state.utf8_pending` for the next token. That part of the design works.

For bytes that are present and simply malformed it throws, at three sites:

- `:425` invalid leading byte
- `:432` invalid continuation byte
- `:438` invalid codepoint (overlong, surrogate, out of range)

In an earlier comment I said I had not traced where that escapes into engine shutdown and
did not want to guess at your lifecycle. I have since traced it, at `6e8b2e2a`:

```
preview_model -> feed_token_bytes -> valid_utf8_prefix_size THROWS
  -> uncaught through EngineCore::commit_pending   (engine_core.h:995-996)
  -> run_decode_round
  -> worker_loop catch-all                          (engine_core.h:1732-1753)
  -> fail_all_locked                                (engine_core.h:1842-1850)
```

`fail_all_locked` sets `failed_ = true` (`engine_core.h:181-183`), tears down every
[...]

PR #161, opened 2026-09-02
fix(frontend): contain malformed UTF-8 in the request instead of terminating the engine

kitaekatt:
Fixes the engine-termination half of #154.

Requires input from the maintainer on the correct resolution (3 options presented). This PR implements one of them
(per-byte U+FFFD) but I can adjust.

## What it changes

`valid_utf8_prefix_size` threw `std::invalid_argument` on a malformed leading byte,
continuation byte or codepoint. The throw was not request-local: it escaped
`commit_pending`, reached `worker_loop`'s catch-all, and ran `fail_all_locked`, which sets
`failed_` and leaves the process serving HTTP while rejecting every later request with
`service_unavailable`. One malformed byte ended service for every client until restart.

The function now returns a `malformed` flag instead, and `feed_token_bytes` consumes the
offending lead byte, publishes U+FFFD and continues.

Cross-token buffering is unchanged: a sequence whose declared length runs past the buffer
may simply be split across tokens, so it is still retained for the next token.

## Overrun branch validation

A truncated sequence and a malformed one can look identical from the lead byte alone when
the declared length overruns the buffer.  The decoder can end up waiting for continuation bytes that never
arrived and `terminalize` then clears `utf8_pending` wholesale — **silently dropping the
valid text that followed the bad byte.**

So the overrun branch now validates the continuation bytes that *are* present. Any byte
failing `(b & 0xC0) == 0x80` proves the sequence malformed immediately, since no later
token can repair it. Genuine truncation still waits.

`test_utf8_and_hidden_eos` catches this precisely — its continuation case asserts
`"x" + U+FFFD + "x"`, and the naive version produces `"x" + U+FFFD`.

## Verification

- **Deterministic reproduction.** Since the sampler-side trigger is stochastic (eight
  seeded attempts with the original `frequency_penalty` recipe did *not* reproduce it), I
  built a decoder-level one: an artifact copy with two equal-width tokenizer vocabulary
  values transposed, so an ordinarily-emitted token decodes to `0xC0`. Weights and offsets
  untouched. Details are in #154.
  - On master: trigger logs `error invalid UTF-8 leading byte`, and the **next ordinary
    request is rejected 503 `service_unavailable`**.
  - With this branch: trigger completes (`finish=length`, 8192 tokens, three U+FFFD), and
    the follow-up returns normally.
  - Same flags, same request JSON, same artifact; the only difference is the binary.
- **`eval/run_qwen3_8_27b_nvfp4_reasoning.sh text` ran to completion on this branch** —
  4/4 suites, 9 h 30 m, zero `engine is unavailable` and zero decode errors. An earlier
  attempt on master died partway through IFBench, which is what prompted #154's severity
  correction. Scores matched the published figures for the NVFP4 artifact (AIME25 96.67%,
  GPQA-Diamond 90.40% exactly; AIME26 93.33% vs 96.67%, one problem at n=30).

## Notes

- `valid_utf8_prefix_size` is non-throwing now and could be `noexcept`.
- The replacement loop's repeated `erase(0, 1)` is O(n²) on a long malformed run.
- This touches the shared Qwen frontend, so it affects the 35B-A3B route too; I have
  exercised it only on 27B NVFP4.

### DIFF
--- a/src/targets/qwen3_6/impl/frontend/frontend.cpp
+++ b/src/targets/qwen3_6/impl/frontend/frontend.cpp
@@ -37,6 +37,7 @@ namespace fi = frontend_internal;
 
 constexpr std::size_t kPatchFeatures   = 1536;
 constexpr std::string_view kThinkClose = "</think>";
+constexpr std::string_view kReplacementCharacter = "\xef\xbf\xbd";
 constexpr std::string_view kThinkingControl =
     "\n\n Considering the limited time by the user, I have to give the solution based on the "
     "thinking directly now.\n</think>\n\n";
@@ -399,7 +400,12 @@ void append_delta(PublishedOutput& output, OutputChannel channel, std::string te
     }
 }
 
-std::size_t valid_utf8_prefix_size(std::string_view bytes) {
+struct Utf8Prefix {
+    std::size_t size = 0;
+    bool malformed    = false;
+};
+
+Utf8Prefix valid_utf8_prefix_size(std::string_view bytes) {
     std::size_t offset = 0;
     while (offset < bytes.size()) {
         const auto lead         = static_cast<unsigned char>(bytes[offset]);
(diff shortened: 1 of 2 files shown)


kitaekatt:
Closing this — it is redundant, and your version is better.

I had been working from a stale `master` (`6e8b2e2a`) and did not fetch before branching, so
I never saw `4cece118`. Everything I reported was true of the tree I was on, but the fix had
been in place upstream since 29 August. That one is on me.

Having now read `consume_generated_utf8` on current `master`, it is better than this PR in
three ways worth naming, so this is not just politeness:

- **It handles a case this PR got wrong twice.** When the declared length overruns the
  buffer, your overrun check sits *inside* the continuation loop, after the bytes that are
  present have been validated. So a `0xE4` followed by `x` is recognised as malformed
  immediately rather than being mistaken for a truncated sequence. My first attempt waited
  for continuation bytes that never arrived and then dropped the buffered text at
  terminalization — silently losing the valid characters after the bad byte. I caught that
  with your own `test_utf8_and_hidden_eos` and fixed it; you had simply never had the bug.
- **It replaces per maximal subpart**, with the comment explaining that the first byte which
  cannot continue the sequence is left for the next iteration. That is the WHATWG behaviour,
  and it is the option I listed as (2) in #154 while asking which you preferred — you had
  already chosen it before I asked.
- **Its lead-byte ranges are stricter than mine.** `0xc2..0xdf` excludes the overlong
  `0xC0`/`0xC1`, and the `0xe0` / `0xed` / `0xf0` / `0xf4` special cases are handled. My
  version accepts sequences yours correctly rejects.

Thank you also for `5f6d44e4` and `5973313d`. The health fix removes a real operational trap
— I had lost about fifteen minutes to an engine that had already died while `/health` kept
answering `ok`, and had been carrying a local workaround that issues a one-token completion
just to tell a dead engine from a busy one. That workaround can now go. And with the fixtures
self-contained I can finally run the frontend suite, which is what caught my own bug above,
so the value of that change was immediate.

Sorry for the noise on #154 — I edited the description to lay the three options out clearly
without noticing you had already closed it.

PR #173, opened 2026-09-04
feat(kv): rk2v4-e8 compressed-KV (E8-root, 208 B/head-token) on the paged-KV engine

danielfparkernz:
## What

Re-port of the rk* compressed-KV storage format — specifically **rk2v4-e8** (E8-root Conway-Sloane lattice Key codes, 2-bit Key / 4-bit Value, 208 B per head-token) — onto the current paged-KV `dev` engine, as a first-class `KvCacheStorage` profile selectable via `--kv-dtype rk2v4-e8`.

This is the successor to [PR #35](https://github.com/Neroued/ninfer/pull/35), which was written against the pre-paged-KV `gqa_attention_*` kernels and closed when the engine moved to the `KVPageGeometry` / `causal_softmax_attention` plane model. This port re-derives the same density win on the current kernel architecture, using the per-format kernel-file pattern (mirroring the K8V4 family) rather than the old `KvCacheCodec` template fold.

## Why it matters on 24 GB cards

NInfer targets a single RTX 5090. The 24 GB **laptop** 5090 (sm_120a, 95 W) is the class where KV-cache headroom is binding, because dense 27B models leave only ~3.4 GiB free after weights. The existing compressed-KV profile, **k8v4** (402 B/head-token), is coarser than rk2v4-e8 (208 B/head-token, **1.93× denser**). All figures below were measured on a 24 GB RTX 5090 Laptop, CUDA 13.1.80, sm_120a — not projections.

### Boot: KV headroom at 256K (Qwen3.6-35B-A3B, 19.6 GiB weights)

| profile | `--kv-capacity 262144` @ 256K |
|-|-|
| k8v4 | FATALs in `auto` mode; explicit mode boots with **574.6 MiB free** |
| **rk2v4-e8** | boots with **1.56 GiB free** (2.7× the margin) |

### Qwen3.8-27B NVFP4 (19.0 GiB weights) — 256K does not fit either

k8v4-class KV wants ~3.64 GiB at 256K but only 3.39 GiB is free after weights, so a dense 27B NVFP4 cannot reach 256K at k8v4's density. rk2v4-e8's 208 B/head-token brings the 256K reservation down enough to **boot at 237,568 tokens**, and to run 160K **with MTP speculative decoding** (`--spec mtp --draft-tokens 3 --lm-head-draft`), whose draft head needs ~1 GiB.

### Fidelity (needle-in-haystack, case-insensitive, 3 needles at 10/50/90% depth)

| | result |
|-|-|
| E8 codec oracle (`tools/test_kv`, 1M tokens) | 5/5 needle retrieval, exact |
| rk2v4-e8 @ A3B, 100K / 150K / 200K | 3/3 each, no OOM |
| rk2v4-e8 @ Qwen3.8-27B groupwise, 219K | 3/3 |
| decode regression | none — 21.8 tok/s baseline matches the groupwise reference; 47.1 tok/s @100K with MTP |

## How

- `include/ninfer/types.h`: add `KvCacheStorage::Rk2v4E8`.
- `src/core/paged_kv_storage.h` `paged_kv_storage_layout()`: Rk2v4E8 geometry — K = {I8, 64, FP16, 4}, V = {U8, 128, FP16, 4} = 208 B/head/token (head_dim 256). `plan_cache()` derives the 4 planes automatically.
- `src/ops/kernel/{e8_lattice.cuh,e8_root_codec.cuh}`: self-contained E8 Conway-Sloane codec cores (port from the 4090 fork; oracle-verified).
- `src/ops/kv_cache/{rk2v4e8_codec.cuh,append/rk2v4e8_kernel.cuh,append/rk2v4e8_launch.cu}`: codec + append kernel family (mirrors `k8v4`).
- `src/ops/softmax_attention/dense/causal_cache/{small_t_rk2v4e8,prompt_rk2v4e8}.{cuh,cu}`: split-KV small-T and causal prompt attention families (mirror k8v4).
- Dispatch: `KvCacheStorage::Rk2v4E8` branches wherever `Fp8KeyNvfp4Value` (k8v4) is routed — append, small-T (both cached and uncached), prompt, `fp32_acc` list, split-capacity bucket.
- `src/serve/serve_options.cpp`: parse `--kv-dtype rk2v4-e8`; capacity/request logs name the profile.
- `tools/test_kv/`: standalone oracle + 1M retrieval verifier (raw-nvcc buildable, no engine deps).

The V inverse-rotation is required (the PV output lands in rotated-V space); gated on `cache.storage == Rk2v4E8` at the output launch.

## Verified on Beast (RTX 5090 Laptop, 24 GB, sm_120a, CUDA 13.1.80)

- Rebased onto `origin/dev` `a140e7ae` — clean rebase, 6 commits, ~4K insertions, **build 321/321 exit 0**.
- A3B 256K rk2v4-e8: `runtime 1.20 GiB | free 1.59 GiB`, needle 3/3 @100K.
- Qwen3.8-27B (groupwise + NVFP4): 256K-class context + MTP as above.

## Test plan

1. `cmake --build build --target ninfer_kv_e8_verify && ./build/tools/test_kv/ninfer_kv_e8_verify 1000000` → 5/5.
2. Boot a 27B or A3B at `--kv-dtype rk2v4-e8 --max-context 262144 --kv-capacity 262144`; confirm the `capacity` log line names rk2v4-e8 and reports the expected pages/runtime.
3. Needle-in-haystack at 100K (harness in this PR's description / `tools/test_kv/`) → 3/3.

Supersedes PR #35 (which targeted the pre-paged-KV engine).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,9 @@ endif()
 if(BUILD_TESTING)
   enable_testing()
   add_subdirectory(tests)
+  # Standalone E8 compressed-KV codec oracle (no engine / weight deps);
+  # registered so `ctest -R ninfer_kv_e8_verify` exercises the rk2v4-e8 codec.
+  add_subdirectory(tools/test_kv)
 endif()
 if(NINFER_BUILD_BENCHMARKS)
   add_subdirectory(bench)
--- a/PORT-RK2V4E8.md
+++ b/PORT-RK2V4E8.md
@@ -1,0 +1,155 @@
+# Task: port rk2v4-e8 (E8-root compressed KV) kernels onto the new paged-KV base
+
+## Sanity check first
+Run `pwd` and `git -C /tmp/wt-rkport log --oneline -2` — you must be working in
+`/tmp/wt-rkport`, branch `feat/rk-compressed-kv-on-6e2786c5`, HEAD `d465fd63`
+(or a new commit you made on that branch). If HEAD is `6e2786c5` or something
+else, STOP and report — the foundation commit is missing.
+
+## What already exists (DO NOT redo)
+Commit d465fd63 already added:
+- `KvCacheStorage::Rk2v4E8` enum case in `include/ninfer/types.h`
+- Its geometry in `src/core/paged_kv_storage.h` `paged_kv_storage_layout()`:
+  K = {I8, 64, FP16, 4}, V = {U8, 128, FP16, 4}  (208 B/head/token; head_dim must be 256)
+- serve `--kv-dtype rk2v4-e8` parsing in `src/serve/serve_options.cpp`
+- E8 codec cores (self-contained, oracle-verified 5/5): `src/ops/kernel/e8_lattice.cuh`,
+  `src/ops/kernel/e8_root_codec.cuh`, plus `tools/test_kv/` (oracle already passes:
+  /tmp/e8_verify exists, do not rebuild unless needed)
+
+## Your job
+Wire the rk2v4-e8 kernels so `--kv-dtype rk2v4-e8` runs end to end. The new base
+ABANDONS the old template-parameter approach and uses one kernel file family per
+format. Mirror the K8V4 family exactly:
+
+Create (mirror these files, substituting the E8-root codec):
+1. `src/ops/kv_cache/append/rk2v4e8_kernel.cuh` + `rk2v4e8_launch.cu`
+   (mirror `k8v4_kernel.cuh` / `k8v4_launch.cu`)
+2. `src/ops/softmax_attention/dense/causal_cache/small_t_rk2v4e8.cuh` + `.cu`
+   (mirror `small_t_k8v4.cuh` / `small_t_k8v4.cu`)
+3. `src/ops/softmax_attention/dense/causal_cache/prompt_rk2v4e8.cuh` + `.cu`
+   (mirror `prompt_k8v4.cuh` / `prompt_k8v4.cu`)
+4. Declare the new launch functions in `src/ops/softmax_attention/dense/causal_cache/launch.h`
+   and `src/ops/kv_cache/append/launch.h` (next to the k8v4 ones)
+5. Add the new .cu files to `src/CMakeLists.txt` where the k8v4 ones are listed
+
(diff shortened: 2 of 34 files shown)


danielfparkernz:
Addressed all three in `daf60f36`:

**P1 — wire rk2v4-e8 through every product parser (fixed).** The profile was only parsed in `ninfer-serve`; `apps/cli/options.cpp`, `apps/perplexity/main.cpp`, and `bench/.../ninfer_bench_support.cpp` rejected it. Added the parse, the enum→name mapping, the usage text, and the error string to each, mirroring the serve parser (canonical name `rk2v4-e8`). Added `rk2v4-e8` parse + report-name assertions to `test_cli_options` and `test_ninfer_bench_support`; verified both pass, the perplexity binary now accepts `--kv-dtype rk2v4-e8` (advances past the dtype check to the corpus error), and CLI/bench `--help` list the profile.

**P2 — register the documented verification target (fixed).** `tools/test_kv` is now wired into the root `CMakeLists.txt` under `BUILD_TESTING`, so the documented top-level build and `ctest -R ninfer_kv_e8_verify` work. `ninfer_kv_e8_verify` now builds and runs as a real target (1M-token needle retrieval passes 5/5).

**P2 — "tests separate formats, not the production codec" (scoping corrected, not a code change).** You're right that the oracle does not exercise the production route: `test_kv/test_e8_codec.cuh` is a deliberately self-contained reimplementation of the E8 Conway-Sloane lattice / 240-root math — it does not `#include` `src/ops/kernel/e8_*.cuh`. So the 5/5 validates the *mathematics*, not the production `rk2v4e8_codec.cuh` 208 B plane layout, the V Hadamard rotation/inverse, or paged addressing. I corrected the `CMakeLists.txt` / `README.md` to state that scope explicitly (and why the oracle is kept independent: a self-including test would be circular). The production route — 208 B representation, V rotation, paging, and final attention output — is validated by the end-to-end boot + needle-in-haystack verification recorded in `PORT-RK2V4E8.md`. Happy to add a direct production-codec unit test (e.g. encode a known K/V plane, compare against a reference reconstruction) if you'd like that folded in too — flagging that as the one genuinely useful follow-up here.

PR #220, opened 2026-09-09
The fused GDN convolution epilogue carries the unrounded projection, so a restart from the published state does not reproduce the next column

MichaelDementii:
`GdnConvEpilogue::store` publishes the represented column and then carries the raw one:

```cpp
// src/ops/gdn_input_proj/gdn_conv.cuh:113-116
publish.publish(token, batch_row, row, s1, s2, p);   // stores __float2bfloat16_rn(p)
s2 = p;                                              // carries the raw accumulator
```

The window a call starts from is read out of BF16 (`gdn_conv.cuh:76-78`), so column 0 uses a
represented window and every later column does not. Restarting the Op from the state it published
for column `j-1` therefore does not reproduce column `j`.

`docs/maintainer/replayssm-gdn.md:496` makes the verify→fold commit an exact gather *provided* the
record is the same BF16 represented column the baseline writes to history;
`qwen3.6-35b-a3b-model.md:114` describes the record row the same way. Fold replays from a window one
rounding step away from the one verify walked.

## Change

One line: round the carry to the BF16 that was just published. The accumulator, `publish`, and the
bytes written to history and to `conv_record` are untouched.

## Evidence

New Op test: restart from the published slot and compare bit for bit, on both publish forms, at
`T` = 2, 3, 4, 8, 16.

| | `b88c0f6` | with this change |
|---|---:|---:|
| restart mismatches | **168** | **0** |
| `T=17`, materialized route (control) | 0 | 0 |
| `z` / `conv_record`, bare projection (control) | identical | identical |
| `ctest` | 114/114 | 116/116 |

The controls place the fault: a column's projection does not depend on the position it occupies in
the call, and the route that stores the projection through BF16 already satisfies the invariant —
so the two registered routes of one Op disagreed.

## What changes for a user, and what does not

Speculative output changes on most prompts; this is a semantic fix, not a no-op. It does **not** buy
speculative/non-speculative parity: over 6 prompts × 3 KV formats the count of prompts whose MTP
output differs from non-speculative goes 4/6 → 5/6 (int8), 5/6 → 6/6 (fp8), 5/6 → 4/6 (bf16).

Draft acceptance over those 18 cells goes 72.2% → 73.3% (9 up, 5 down, 4 unchanged); the arms
produce different text, so that is "acceptance does not collapse", not a speedup. No quality
benchmark was run, so nothing is claimed about output quality. The argument is the invariant.

Scope: 1 line of product code, 1 new test (+353), 1 line of `tests/CMakeLists.txt`.

### DIFF
--- a/src/ops/gdn_input_proj/gdn_conv.cuh
+++ b/src/ops/gdn_input_proj/gdn_conv.cuh
@@ -113,7 +113,9 @@ struct GdnConvEpilogue {
             publish.publish(token, batch_row, row, s1, s2, p);
             s0 = s1;
             s1 = s2;
-            s2 = p;
+            // Carry forward the column that the next column would read back, not the wider
+            // accumulator: the published window is a sequence of BF16 represented columns.
+            s2 = __bfloat162float(__float2bfloat16_rn(p));
         }
     }
 };
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -350,6 +350,9 @@ ninfer_add_op_test(ninfer_gdn_input_proj_conv_snapshot_test
 ninfer_add_op_test(ninfer_gdn_input_proj_conv_record_test
   SOURCES ops/test_gdn_input_proj_conv_record.cpp
   LIBRARIES ninfer_ops)
+ninfer_add_op_test(ninfer_gdn_input_proj_conv_restart_test
+  SOURCES ops/test_gdn_input_proj_conv_restart.cpp
+  LIBRARIES ninfer_ops)
 ninfer_add_op_test(ninfer_gated_delta_net_replay_record_test
   SOURCES ops/test_gated_delta_net_replay_record.cpp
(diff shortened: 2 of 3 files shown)


Neroued:
This is not a numerical requirement: the Op contract explicitly allows private convolution intermediates to remain unrounded. Exact equivalence is required only for ReplaySSM record/fold versus the corresponding full computation, which existing tests already cover.

The reported restart differences do not violate that contract and do not justify introducing additional rounding.

Please also keep PR descriptions concise and focused on the final change:

- Start with the existing contract being violated, or the concrete optimization being proposed.
- Explain what changes and why it is needed.
- Summarize relevant validation and limitations; put detailed measurements and logs in a separate report.
- Update the description when the implementation changes. Remove superseded claims and revision history.
- Avoid repeating conclusions or presenting an assumed invariant as an established requirement.

A reviewer should be able to identify the problem, the change, and the supporting evidence without reconstructing the investigation.

PR #221, opened 2026-09-09
MTP graph profiles carry no topology class, so one executable serves two attention routes and any draft window past five fails at startup

MichaelDementii:
`instantiate_graph_family` builds one `cudaGraphExec_t` per topology class and installs every other
profile of that class through `cudaGraphExecUpdate`, which cannot cross a change of node count.

`dflash_graph_profiles` honours that twice: it breaks the frontier where the attention route flips
(`variant.cpp:49-51`) and derives `topology_class` from the same predicate (`:114-124`).
`mtp_graph_profiles`, three functions above it in the same file, does neither — every profile it
returns carries class 0.

## It is a hard failure, not a style point

Raising the MTP draft cap — four constants, no other change — on this build:

| draft tokens | master | with this change |
|---|---|---|
| 1–5 | ok | ok |
| **6, 7, 8, 11, 12, 15** | `error: startup failed \| preparing CUDA graphs`<br>`cudaErrorGraphExecUpdateFailure (update result 2)` | ok |

## Change

`mtp_graph_profiles` now breaks the frontier at the route flip and sets `topology_class` from the
same predicate `dflash_graph_profiles` uses.

* the predicate is renamed `verify_uses_chunked_small_t` because both planners read it now — a
  second copy is what would drift;
* it gains the width bound the real route has: `causal_attention_resolve_route` returns `Prompt` for
  every envelope past `kMaximumVerifyTokens`, so claiming a target dependence beyond width 16 costs
  a topology class no verify path can request;
* `verify_route_flip_target` locates the boundary by bisecting that predicate — monotone in
  `max_visible_keys` — rather than restating the route table, which is how the planner and the route
  could drift apart in the first place.

1 file, +46 / −4.

## Evidence

New contract test over capacities 2048 / 16384 / 65536 / 262144, draft windows 1..16, all five
`KvCacheStorage` values: no profile may span a route flip inside its own frontier range, and no
topology class may carry two routes.

| | `b88c0f6` | with this change |
|---|---:|---:|
| contract violations | **1150**, over 40 (capacity, k) pairs, k = 6..15 | **0** |
| `ctest` | 114/114 | 116/116 |

## At the shipped cap it is a no-op

At k ≤ 5 every width resolves to `SmallT`, so no class changes and the added boundary sits behind
`if (flip_target != 0)` — the profile list does not move by one element. Measured too: stock caps,
greedy, k = 1..5, two binaries of different `md5` print byte-identical token ids, 5 of 5, while the
same comparison separates k=1 from k=5.

`dflash_graph_profiles` is unchanged: the renamed predicate takes the same arguments, and the new
width bound cannot fire there because DFlash caps at 15.

This does not raise the MTP draft window and proposes no wider window — that needs its own case on
a workload. It removes the crash that meets anyone who raises it, and pins the contract with a test.

### DIFF
--- a/src/targets/qwen3_6_35b_a3b/impl/variant.cpp
+++ b/src/targets/qwen3_6_35b_a3b/impl/variant.cpp
@@ -7,6 +7,7 @@
 #include "ninfer/ops/sparse_moe.h"
 
 #include <algorithm>
+#include <limits>
 #include <stdexcept>
 
 #define NINFER_QWEN36_VARIANT    ::ninfer::targets::qwen3_6_35b_a3b::detail::Variant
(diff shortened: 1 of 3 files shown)


MichaelDementii:
On the Codex P1: it is right that nothing reaches this today. `kMtpDecodeMaximumDrafts` is 5, so
`verify_route_flip_target` returns zero for every accepted window and every class stays 0 — which is
also why the PR shows byte-identical output at k = 1..5 rather than claiming a behaviour change.

What I would put on the other side, for you to weigh:

- `dflash_graph_profiles` carries exactly this guard, three functions below, for a window that *is*
  reachable. The two planners consume one route table and only one of them mirrors it.
- The failure mode is not a wrong answer but `cudaErrorGraphExecUpdateFailure` at startup, and the
  test is the only thing that would catch a reintroduction.
- It is 46 lines and it is not on any hot path.

If you would rather the MTP limit stay where it is and the planner not carry a guard for a window it
cannot be asked for, say so and I will close this — the report itself is in the description and
loses nothing by the PR going away.

Issue #223, opened 2026-09-09
Named `tool_choice` is rejected as unenforceable, but its constraint is prompt text


MichaelDementii:
## Observation

Every forced form of `tool_choice` is refused with HTTP 400 on all three APIs:

| request | error code |
|---|---|
| Chat `tool_choice: "required"` | `tool_choice_not_supported` |
| Chat `tool_choice: {"type":"function","function":{"name":"get_weather"}}` | `tool_choice_not_supported` |
| Chat `allowed_tools` with `mode: "required"` | `tool_choice_not_supported` |
| Responses `tool_choice: "required"` and named | `tool_choice_not_supported` |
| Messages `tool_choice: {"type":"any"}` and `{"type":"tool"}` | `invalid_request_error` |

```
tool_choice='required' requires at least one tool call, which NInfer cannot guarantee
```

`docs/serving.md` states the reason once for the whole group: NInfer does not "enforce JSON Schema
through constrained decoding", so `strict:true`, required and named tool choice, hosted tools and
custom tools are all rejected together.

## Why it matters

`auto` is the only executable value, and it does not stand in for the others. Three tools
declared, 16 requests of the kind a client sends when its own contract has already decided that a
call must happen:

* `tool_choice: auto` -> a tool call in **1 of 16**.

A client that needs a call has no supported way to ask for one. The request fails instead.

## What a named choice actually needs

Qwen's call syntax opens with a literal that NInfer itself renders today
(`chat_template.cpp`, `render_tool_call`):

```
<tool_call>
<function=NAME>
```

For a **named** choice that literal is the entire constraint, and it is prompt text. Nothing has
to be constrained in the sampler. NInfer already continues a partially written assistant turn
(`PromptContinuationMode::ContinueFinalAssistant`), so the mechanism can be measured from outside
today, through an Anthropic assistant prefill. Same 16 prompts, each forced to each of the three
tools, 48 requests:

| forced span | function name | `<tool_call>` block closed |
|---|---|---|
| `<tool_call>\n<function=NAME>\n` (named choice) | **48/48**, by construction | **42/48** |
| `<tool_call>\n<function=` (the `required` analogue) | 13/16 declared | 15/16 |

The two rows point in opposite directions, and that is the point of this report:

* **`required` over several tools is genuinely not enforceable this way.** The name is still
  sampled, and the model produced `tool_list`, `do_nothing` and `public.get_weather`. The current
  rejection is correct and should stay.
[...]

PR #224, opened 2026-09-09
Execute a named tool_choice by opening the call in the generation prompt

MichaelDementii:
Implements #223.

**Scope is not confirmed yet.** #223 ends with three open questions, and this branch is the
concrete answer I would give to them, not a change built on an agreed design. If any answer
differs, I will rework or close it; the Issue stands on its own either way.

## What changes

A forced `tool_choice` is executed by ending the generation prompt with that call's opener:

```
<|im_start|>assistant
<think>

</think>

<tool_call>
<function=send_email>
```

The answer can only continue inside that call. Nothing in the sampler changes.

Accepted, because the opener determines the function:

| API | accepted forms |
|---|---|
| Chat Completions | `{"type":"function",...}`; `"required"` over a single callable tool; `allowed_tools` with `mode:"required"` narrowed to one |
| Responses | `{"type":"function","name":...}`, namespaced identity included; `"required"`; `allowed_tools` `mode:"required"` |
| Messages | `{"type":"tool","name":...}`; `{"type":"any"}` over a single callable tool |

Still rejected, each with a message that says which guarantee is missing:

- `required` / `any` over several callable tools. The function itself would still be sampled, and
  that is the part no prompt can constrain: #223 measured 13/16 declared names when only
  `<tool_call>\n<function=` is forced.
- A forced choice on a reasoning-enabled request. That prompt ends inside `<think>`, where the
  opener has no position.
- A name absent from the declared tools, which today reaches the model as an unenforced hint.
- `strict:true`, `parallel_tool_calls:false`, hosted and custom tools: untouched.

## Design

Three pieces, each on the side that owns the fact.

1. **Adapters** resolve every accepted form to one function name in `ToolChoice::forced_name`.
   `required` and `any` over a single callable tool are the same span as a named choice, so they
   share one helper rather than a second code path.
2. **Chat template** appends the opener after the generation prompt. It is the literal
   `render_tool_call` already writes into assistant history, now named `kToolCallOpen` once instead
   of spelled out twice. `render` rejects the three positions the opener cannot occupy: a
   continuation, a closed turn, and a thinking prompt.
3. **Output decoder** is seeded with the same literal, because the model never emits it. That is
   the whole of the output-side change: the region the parser already builds simply starts where
   the prompt left off.

`resolve_prompt_semantics` rejects the reasoning-enabled combination next to the existing
assistant-prefill rule, so the two prompt shapes that cannot carry an opener are refused in one
place.

One addition beyond seeding: a turn that ends on a closed `</function>` has its outer
`</tool_call>` supplied, and `forced_call_closed` records that it happened. Only the outer tag is
ever supplied, so no argument byte is invented; anything less complete still falls back to verbatim
text exactly as today. This answers question 2 of the Issue, and it is the difference between a
guarantee and a strong best effort. It is easy to drop if you would rather not own it.

## Contract and observability

- `docs/serving.md`: the four places that describe `tool_choice` now state what is executable, the
  reasoning-disabled requirement, and that the forced call is present and first while a model that
  keeps writing may add further calls, exactly as under `auto` with parallel calls enabled.
- Request log: `tool_choice` is recorded as `function:<name>`, `forced_call_closed` joins the
  tool-call diagnostics, schema version 20 -> 21.
- `tool_choice` `auto` and `none` render and parse exactly as before; the seeding and the
  completion are both reachable only through a forced name.

## Verification

Environment: this branch on `b88c0f6f`; artifact `qwen3_6_35b_a3b.ninfer`; RTX 5090 32 GiB, driver
591.86, CUDA 13.1, Ubuntu 26.04 under WSL2, gcc 15, Release.

```
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON \
      -DNINFER_BUILD_APPS=ON -DCMAKE_CUDA_ARCHITECTURES=120a
cmake --build build -j 28
ctest --test-dir build -j 8
NINFER_QWEN3_6_35B_A3B_WEIGHTS=/root/models/qwen3_6_35b_a3b.ninfer \
  ctest --test-dir build -R 35b
```

- `ctest` **114/114**, plus the three 35B artifact tests **3/3** with the weights variable set.
- New coverage: `test_frontend.cpp` asserts the generation prompt ends with the opener, that the
  three illegal positions throw, and the three decoder outcomes (complete call, call closed by the
  decoder with `forced_call_closed`, unfinished argument falling back to text). The three schema
  tests assert the accepted forms resolve a `forced_name`, that an undeclared name and a
  multi-tool `required` are refused, and that a forced choice with reasoning enabled is refused.

End to end, `ninfer-serve <artifact> --port 18080 --max-context 32768 --greedy`, reasoning disabled
per request, the same 16 prompts and 3 tools as #223, one deterministic run each:

| | result |
|---|---|
| `tool_choice: auto`, 16 requests | a tool call in **1/16** |
| `tool_choice: {"type":"tool","name":N}`, 3 tools x 16 prompts | the named call in **48/48** |

47 of the 48 returned that call alone; one also carried a second call to another declared tool,
which is the documented parallel-call behavior, not a lost guarantee. The 16 prompts were replayed
under each of the three tools with prefix reuse enabled by default, and each replay returned its
own tool, so the forced suffix is not served from a cached prompt. A streamed Chat request with a
named choice assembles `get_weather` / `{"city":"New York"}` across 3 SSE frames and finishes
`tool_calls`.

The decoder's closing completion did **not** fire in those 48 requests: every forced call closed on
its own. It is exercised by the decoder test, and #223 measured the same model ending 6 of 48 turns
on `</function>` under an assistant-prefill emulation of the same opener, so the case is real
rather than hypothetical.

## Not verified, and what it costs

- No performance claim is made and no benchmark was run. The prompt grows by the opener and the
  request otherwise executes as before; if you want a number for that, say which bench.
- Only the 35B-A3B artifact was exercised. The other registered templates were not, so a template
  whose tool syntax differs would need its own check before this is claimed for it.
- Reasoning-enabled requests are rejected rather than supported. The Issue sketches the
  control-span route (`OutputSession::pending_control_tokens`) for that case; none of it is
  implemented here, and it is the larger half of the use case.
- Vision, `dflash`, and MTP paths were not exercised with a forced choice. The change is prompt and
  decoder only and does not touch the round, but that is reasoning, not measurement.

I understand and can defend the complete diff, and I will follow up on review here.


MichaelDementii:
@codex review

MichaelDementii:
Both taken, in `c0e5d099`.

**P1, the log consumer.** You are right and it would have broken every campaign:
`require_server_log_identity` compares the triple exactly, so with the producer at 21 and
`run_serve_corpus.py` at 20 the very first `server_start` raises. Both the consumer and the two
places in `tests/test_serve_corpus.py` that codify the current version now say 21. Control: with
the consumer alone put back to 20, `pytest tests/test_serve_corpus.py` fails **2 of 4**; restored,
**4 pass**.

**P2, the forced-call seed.** Also right, and the fix is to delete the second copy rather than
reconcile it. `OutputOptions::forced_tool_name` is gone; the name now lives in the
`ToolCallOutputContract` that the prepared prompt already carries and hands to the output session.
That is the only object that knows which opener the prompt actually ends with, so the two can no
longer disagree — not by validation, but because there is nothing left to disagree with. The
decoder's separate `forced_tool_name` parameter is gone with it; it reads the contract.

`Engine::prepare` and `Engine::submit` stay separate calls and the public surface got smaller: the
protocol adapter sets the name once, on `PromptInput`. The forced-call test in
`tests/targets/qwen3_6/test_frontend.cpp` now passes an `OutputOptions` with no name at all and
still sees the seeded opener complete the call — which is the regression evidence that the seed
travels with the prompt.

`ninfer_qwen3_6_frontend_test`, `ninfer_openai_schema_test`, `ninfer_anthropic_schema_test`,
`ninfer_openai_responses_test` all pass; `clang-format` clean on the six C++ files.

@codex review

MichaelDementii:
Both taken, in `bc9502c0`.

**The opener in fallback content.** You are right, and the old test permitted it by asking only
that content be nonempty. The decoder now records how many bytes of the region came from the prompt
and skips them on the fallback path. That is safe for exactly the reason the forced turn is
special: `saw_tool_marker_` is set in the constructor, so `feed()` always takes the append branch
and never rebuilds the region — the seeded prefix stays at its head for the life of the session.

The truncated test now asserts origin rather than length: the content must not contain
`<tool_call>` or `<function=TaskUpdate>` and must begin with `\n<parameter=taskId>`, which is the
first byte the model actually produced. Control: with the fallback put back to returning the whole
region and nothing else changed, that check fails; restored, the suite passes.

**The documentation.** Also right — a consumer written from `docs/serving.md:854` or
`tools/bench/README.md:156` would have demanded v20 and rejected every current log. Both now say
v21, alongside the producer, the campaign consumer and its test.

`clang-format` clean.

@codex review

Neroued:
I plan to support forced tool choice through constrained decoding, with explicit handling of tool selection, call structure, and termination. This is planned work, but it is lower priority.

I’m closing this PR because prompt prefilling and decoder-side repair are not the implementation direction I want to maintain for this feature. We can keep #223 open to track the requirement; there is no need to revise this implementation further.

PR #225, opened 2026-09-09
Fold the sigmoid gate into the causal-attention reduce epilogue: operator -8.3% where it folds, +0.31% decode

MichaelDementii:
All three call sites in `TextContext` issue `causal_softmax_attention` and then `ops::sigmoid_mul`
over the whole output — one more graph node per full-attention layer and per MTP tail.

## Change

`causal_softmax_attention` takes an optional gate. Where the route reaches the shared BF16/INT8
reducer the multiply is folded into its store; every other route — FP8, NVFP4, K8V4, the prompt
kernel, chunked small-T — applies the standalone elementwise kernel inside the Op. One contract
either way, so no caller has to know which route it landed on.

The bytes are the bytes the caller's own `sigmoid_mul` produced. That is a rounding question, not a
formality: the standalone kernel reads what attention *stored*, so it sees the reduce result already
rounded to BF16. The fused epilogue therefore rounds first, widens back, multiplies in FP32 and
rounds to nearest. Keeping the FP32 accumulator would be more accurate and would move tokens.

`causal_softmax_attention_cached` is untouched and keeps the standalone multiply; extending it is a
separate decision.

## Operator

The bench gains `--gate off|standalone|fused`. Across 320 append cells:

| | cells | median kernel time vs `standalone` |
|---|---:|---:|
| route folds the gate (BF16, INT8) | 65 | **−8.33 %**, 58 of 65 improved |
| route does not fold (control) | 255 | **+0.00 %** |

Per storage and geometry: bf16 16/2 −8.33 %, bf16 24/4 −6.67 %, int8 16/2 −8.33 %, int8 24/4
−8.78 %. The 255 unfolded cells are the control that the standalone path is untouched. Timings
quantise to 2.048 µs at these widths, so read the percentages as coarse.

## End to end

--lm-head-draft -r 2 --warmup 1`, metric `decode_output_tok_s_mean`. Ten passes, arm order mirrored
between passes, third arm = master rebuilt under a second label as a zero control.

| | median vs master | range | passes positive |
|---|---:|---|---:|
| this change | **+0.311 %** decode | +0.134 … +0.391 | 10 / 10 |
| zero control | +0.003 % | −0.097 … +0.088 | — |

The worst pass of the change is above the best pass of the control.

**Prefill is not claimed:** −0.06 % median against a zero control of +0.06 % on the same runs.

## Correctness

Each case in `causal_cache.cpp` now runs a second time with a gate and is judged against a third,
ungated run — which separates "the Op repeats itself" from "the gate is exact" instead of conflating
the two. Both checks are exact, on every storage, both geometries, batched and single, graph and
eager.

**Strength control:** dropping the rounding before the multiply — the one subtlety in the epilogue —
makes that check fail. Without it "the test passes" would say nothing.

`ctest` 114/114 on both sides; `clang-format` clean on every touched file.

Scope: 8 files, +224 / −99, of which the bench switch and the test are 129 lines.

### DIFF
--- a/bench/ops/causal_softmax_attention_bench.cu
+++ b/bench/ops/causal_softmax_attention_bench.cu
@@ -108,7 +108,14 @@ struct Result {
     bench::ColdTiming timing;
     std::size_t graph_nodes = 0, workspace_peak = 0;
     int graph_calls = 1;
+    // The mode this row was actually measured in, not the one asked for on the command line: the
+    // cached entry takes no gate parameter, so it is standalone whenever a gate is requested.
+    GateMode gate = GateMode::Off;
 };
+
+const char* gate_name(GateMode gate) noexcept {
+    return gate == GateMode::Off ? "off" : gate == GateMode::Standalone ? "standalone" : "fused";
+}
 
 [[noreturn]] void usage(const char* message) {
     std::fprintf(stderr,
@@ -354,9 +361,9 @@ PagedKVLayerView make_cache_view(DeviceBuffer& k, DeviceBuffer& v, DeviceBuffer&
         .k_pages = Tensor(
             k.p, layout.key.data_dtype,
             {layout.key.data_leading_extent, kPagedKVPageSize, geometry.kv_heads, physical_pages}),
-        .v_pages = Tensor(v.p, layout.value.data_dtype,
-                          {layout.value.data_leading_extent, kPagedKVPageSize, geometry.kv_heads,
-                           physical_pages}),
+        .v_pages       = Tensor(v.p, layout.value.data_dtype,
+                                {layout.value.data_leading_extent, kPagedKVPageSize, geometry.kv_heads,
+                                 physical_pages}),
         .k_scale_pages = layout.key.has_scale()
                              ? Tensor(k_scale.p, layout.key.scale_dtype,
                                       {layout.key.scale_leading_extent, kPagedKVPageSize,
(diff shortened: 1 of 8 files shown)


MichaelDementii:
@codex review

MichaelDementii:
@codex review

MichaelDementii:
Both taken, in `2b7fb9e7`.

**P1, the oracle.** You are right that the two gated checks compared production against
production. The gated output is now judged against the same FP64 attention oracle the ungated
output is judged by, multiplied by a host sigmoid of the gate — an independent path from the
inputs, not a second run of the Op. The parity checks stay as the contract they express: the
folded route must produce the bytes the standalone one produced.

The criterion is the ungated one widened by exactly the second BF16 rounding the gated route adds
(2^-9 on the relative-L2 bound and on the bound relative to the largest reference; the absolute
floor is unchanged). That widening is not cosmetic — with only the L2 bound widened, the oracle
check failed on real cells, e.g. `actual=-0.664062 reference=-0.660504`, which is the per-element
rounding and nothing else. `ninfer_softmax_attention_test` passes on all three suites.

**P2, the CSV.** Each row now carries the gate mode it was measured in, on the console line and in
the CSV, as the **effective** mode rather than the flag: the cached entry takes no gate parameter,
so with `--gate fused` it records `standalone`. From one run:

```
append,d256-h16-kv2,bf16,identity,graph,cold,fused,1,4,...
cached,d256-h16-kv2,bf16,identity,graph,cold,standalone,1,4,...
```

`clang-format` clean on both files.

@codex review

MichaelDementii:
Right — the `--profile` branch returns above that line. `d63848ec` puts the gate in the `PROFILE`
header through the same helper the CSV and the result line now use, so all three name the effective
mode. Four runs, two entries by two modes:

```
PROFILE entry=append ... cache=cold gate=off        B=1 W=4 ...
PROFILE entry=cached ... cache=cold gate=off        B=1 W=4 ...
PROFILE entry=append ... cache=cold gate=fused      B=1 W=4 ...
PROFILE entry=cached ... cache=cold gate=standalone B=1 W=4 ...
```

`clang-format` clean.

@codex review

MichaelDementii:
Closing in favour of #268, rewritten on today's master.

`04350ba9` removed `src/targets/` in favour of `src/models/qwen3_5/`, so the branch could not be rebased - the file its call site lived in no longer exists. The other seven files this change touches are byte-identical between its old base and master, so the Op side ports verbatim; the call site moved to `src/models/qwen3_5/execution/text.cpp`.

Everything was re-measured on `dc58675f`, and two things came out differently from what this PR's body says, both worth stating rather than quietly dropping:

- **The decode number is gone.** It claimed +0.311 % with a zero control. On the machine I have now the identical-baseline arm reads +0.55 % median on decode over a ±0.9 % band, so the stand does not resolve the effect and #268 does not claim one. The operator result stands and is measured in graph nodes as well as in time.
- **The operator claim is narrower and better grounded.** `-8.33 % on 65 cells` becomes -6.19 % on 76, where "folds" is read off the bench's own `graph_nodes` column rather than assumed, and the saving turns out to be one node worth about 2 µs that does not grow with the cell. The cells the product actually issues save less than the band median - 1.03 µs - and #268 says so.

PR #226, opened 2026-09-09
Remove the empty graph nodes of the MTP draft phase: +0.33% decode on Qwen3.6-35B-A3B

MichaelDementii:
Third of three node-count reductions measured in one session; #222 and #225 are the others. All
three are independent mechanisms but all three touch `text_context_impl.h`, so whichever lands first
leaves the other two a trivial rebase.

**Why one PR and not three.** Separately these are +0.090 %, +0.095 % and +0.040 % of decode against
a zero-control band of ±0.15 % on this hardware — none is resolvable alone. Together they are.

## Change

| | was | is |
|---|---|---|
| `ops::argmax` opened its atomic contest with a 4-byte `cudaMemsetAsync` per slice | memset node | initializer kernel |
| the MTP stem normalized embedding and hidden only to lay them side by side | 3 nodes | 1 |
| the MTP tail added the residual only for the next norm to read it back | 2 nodes | 1 |
| the draft chain copied the new hidden state over the old one every step | copy per step | two buffers |

Both new Ops carry a `*_supported()` predicate with the three separate calls as the fallback, so an
unregistered shape keeps working.

## Nodes

One traced run — 8192-token prefill, 64 generated tokens, two repetitions, `nsys
--cuda-graph-trace=node`:

| | master | this |
|---|---:|---:|
| `[CUDA memset]` | 676 | 503 (now kernel nodes) |
| `rmsnorm_cta_bf16x2` | 3132 | 2735 |
| `mtp_pack_fc_input` | 134 | 0 |
| `residual_add_bf16x8` | 132 | 3 |
| `memcpy Device-to-Device` | 100 | 18 |

## End to end

--lm-head-draft -r 2 --warmup 1`, metric `decode_output_tok_s_mean`. Twelve passes, arm order
mirrored, third arm = master rebuilt under a second label as a zero control.

| | median | range | passes positive |
|---|---:|---|---:|
| this change | **+0.334 %** decode | −0.34 … +0.40 | 10 / 12 |
| zero control | −0.027 % | −0.79 … +0.22 | — |

Ten of the twelve sit in a tight band of +0.27…+0.40 %. The two that do not, and the one outlier of
the control, coincide with a build running on the neighbouring container, which shares the host CPU
quota — so separation is **not** full and I am not claiming it. Prefill is unchanged: −0.009 %
median against a control of −0.029 %.

## Correctness

Output byte-identical on the same prompt, greedy, int8 KV, mtp3. `ctest` 114/114. The stem Op test
checks the fused result against the three-Op path it replaces, in the same process on the same
inputs.

Scope: 10 files, +756 / −18, of which 288 are the new Op test.

### DIFF
--- a/include/ninfer/ops/mtp_pack.h
+++ b/include/ninfer/ops/mtp_pack.h
@@ -30,6 +30,67 @@ void mtp_pack_fc_input(const Tensor& embedding_norm, const Tensor& hidden_norm, 
                        cudaStream_t stream);
 
 /**
+ * Op: mtp_norm_pack_fc_input
+ *
+ * Math / indexing:
+ *   out[0:D, t]  = rmsnorm(embedding[:, t], embedding_weight)
+ *   out[D:2D, t] = rmsnorm(hidden[:, t], hidden_weight)
+ *   with the unit-offset weight convention, i.e. the weight used is (1 + w).
+ *
+ * Logical shapes:
+ *   BF16 embedding and hidden [D,T], weights [D], out [2D,T], all contiguous.
+ *
+ * Numeric:
+ *   Identical to rmsnorm() applied to each half followed by mtp_pack_fc_input(): the same block
+ *   width, pair decomposition, reduction and epilogue produce the same bytes. Returns false from
+ *   mtp_norm_pack_fc_input_supported() for shapes it does not cover, and the caller keeps the
+ *   three-Op path there.
+ *
+ * Effects:
+ *   Writes the full output. Inputs, weights and output must not alias.
+ *
+ * Workspace:
+ *   None. The Op has no persistent state side effect.
+ */
+[[nodiscard]] bool mtp_norm_pack_fc_input_supported(const Tensor& embedding,
+                                                    const Tensor& embedding_weight,
+                                                    const Tensor& hidden,
+                                                    const Tensor& hidden_weight, const Tensor& out);
+
+void mtp_norm_pack_fc_input(const Tensor& embedding, const Tensor& embedding_weight,
+                            const Tensor& hidden, const Tensor& hidden_weight, Tensor& out,
+                            float eps, cudaStream_t stream);
+
+/**
+ * Op: mtp_residual_norm
+ *
+ * Math / indexing:
+ *   residual[:, t] += delta[:, t]           (pairwise, FP32 sum, round to nearest BF16)
+ *   out[:, t]       = rmsnorm(residual[:, t], weight)   with the unit-offset weight convention.
+ *
+ * Logical shapes:
+ *   BF16 delta and residual [D,T], weight [D], out [D,T], all contiguous.
+ *
+ * Numeric:
+ *   Identical to residual_add() followed by rmsnorm() on the updated residual: the same pairwise
+ *   FP32 addition and rounding, then the same block width, pair decomposition, reduction and
+ *   epilogue. mtp_residual_norm_supported() reports the shapes this route covers; the caller keeps
+ *   the two-Op path elsewhere.
+ *
+ * Effects:
+ *   Updates the full residual in place and writes the full output. delta, weight and out must not
+ *   alias the residual.
+ *
+ * Workspace:
+ *   None. The Op has no persistent state side effect.
+ */
+[[nodiscard]] bool mtp_residual_norm_supported(const Tensor& delta, const Tensor& residual,
+                                               const Tensor& weight, const Tensor& out);
+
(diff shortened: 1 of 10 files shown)


MichaelDementii:
@codex review

MichaelDementii:
Taken, in `2a9dc0fa`. Both fused Ops are now judged against an independent oracle as well:

* the stem — the naive FP64 RMSNorm oracle and the criterion `tests/ops/test_rmsnorm.cpp` judges
  `ops::rmsnorm` by, evaluated from the represented BF16 inputs and compared against both halves of
  the packed output;
* the tail — the same for the normalised output, and an **exact** oracle for the residual, because
  a BF16 sum of two BF16 values is exact in double.

The parity checks stay where they were. They are not weak evidence, they are the claim these Ops
make: the fused route has to produce the bytes the three-Op path produced, and anything that shifts
a bit shows up as a different draft token.

Strength control, and it answers your point directly. The same defect placed in **both** the fused
kernel and `ops::rmsnorm` (`+ eps` → `+ 0.05f` in `mtp_pack.cuh` and `rmsnorm.cuh`) leaves every
parity check passing and fails only the new ones — 12 of 12 oracle cells, no other failure in the
run:

```
mtp_norm_pack_fc_input D=5120 T=1 against the oracle: reduction criterion failed
  at index 4581 actual=2.4375 reference=2.45659
```

With the kernels restored, `ninfer_mtp_pack_test` is `OK`. `clang-format` clean.

@codex review

MichaelDementii:
Taken, in `03e67bc3`. The route now admits **2048 and 5120** and nothing else; `1536`, `4096` and
`8192` go back to the composed three-Op path, and the `512x8` instantiation, which no admitted
width reaches any more, is gone rather than left dangling.

The same predicate serves the residual-and-norm tail, so that route narrowed with it. The test
pins both directions on both routes — `1536` and `4096` now assert *declined*, `384` still does,
`2048` and `5120` still assert admitted. `ninfer_mtp_pack_test` is `OK`, `clang-format` clean.

Why the ladder was mirrored in the first place is unchanged and still in the comment: an admitted
width has to land on the instantiation `src/ops/launcher/rmsnorm.cu` would have picked for it, or
the sum of squares splits differently and the fused row stops being bit-exact. Narrowing does not
weaken that — it only stops offering it to widths nothing asks for.

@codex review

MichaelDementii:
Right, and fixed in `59f80df2` — both entry points now apply the same check `ops::rmsnorm` applies,
with the same message shape.

Pinned in the test rather than only written: `eps_contract` calls both Ops at both registered
widths with `0`, `-1e-6`, NaN and infinity and requires each to throw. Removing the two checks and
nothing else turns that into **16 failures**, one per call; restoring them returns the suite to
`OK`. `clang-format` clean.

@codex review

MichaelDementii:
Both taken, in `d121079d`.

**The contract.** You are right and the rule is explicit — `op-development.md` §3 forbids freezing
reduction association, launch decomposition and bitwise equality in a contract, and my Numeric
fields froze all three. A legitimate retuning of `ops::rmsnorm` would have broken the contract
without changing what the Op computes. Both fields now state the formula, the positive-and-finite
`eps` domain, the BF16 roundings that are the observable boundaries, and the FP64-oracle criterion
— enough to write an independent implementation, which is what §3 asks for. Bit-parity with the
composed path is still checked, in the tests, as regression evidence rather than as the contract.

**The weight shape.** Also right: `numel()` let `[1,D]` and `[D/2,2]` take the fused route although
`ops::rmsnorm` refuses them. Both Ops now check the whole shape. Pinned: `weight_shape_contract`
feeds those two views to all three weight parameters at both widths; reverting to the element count
and nothing else turns that into **12 failures**, and the suite is `OK` with it restored.

`clang-format` clean on all three files.

@codex review

MichaelDementii:
Closing this one to rework it, and I will resubmit.

It is 916 lines across ten files for **+0.33 % decode**, and that figure is the three mechanisms
bundled: measured separately they read +0.090 %, +0.095 % and +0.040 %, each below this machine's
±0.15 % null band. Only together do they clear it, and even then on 10 passes of 12 rather than with
full separation — which the body says, but saying it does not make the package easier to accept. The
ratio of review surface to demonstrated gain is wrong, and that is my error in packaging, not
something for you to adjudicate.

What I will do instead: split it, lead with whichever mechanism can carry its own measurement on its
own, and drop the ones that cannot. The new-format `mtp_pack` Op and its test are worth keeping on
their own terms; the empty-node removal around them is what has to justify itself separately.

Nothing here is withdrawn as wrong — the output is bitwise identical and `ctest` was 114/114. It is
withdrawn as too large for what it buys.

PR #238, opened 2026-09-14
NVFP4 W4A4 route 1/4: feed the activation scales as one tile instead of BlockM 16-byte requests

MichaelDementii:
Replaces #160, closed to re-prepare the submission. Same mechanism; the description now states the
numbers on current master rather than the pre-`ee9d5192` ones, and the duplicated route predicate that
branch carried is gone.

## Problem

The NVFP4 W4A4 TMA route stages four operands, and one of them is fed at a fraction of the rate of the
others. The activation-scale descriptor is a 16-byte box over a row-major plane, so one pipeline stage
issues `BlockM` separate 16-byte requests to deliver 4 KiB, while the weight codes beside it travel as
one 64-byte box per row. That is a property of the descriptor, not of the data: the plane's row stride
is `kGroupsPerRow` bytes, so a box 16 bytes wide and `BlockM` rows tall is `BlockM` transactions
however the plane is filled.

## Change

The quantizer writes the activation-scale plane tile-contiguous when the GEMM that follows will take
the TMA route, and row-major otherwise, because the small-T and GEMV routes read the same buffer. The
descriptor is reshaped to match and the request coordinate becomes a tile index. Byte order inside a
tile does not change, so the shared image is identical and the MMA-side reader is untouched; only the
producer loop addresses the box differently.

The quantizer that writes the plane and the descriptor that reads it must agree about the layout, so
the change states that agreement once in each place it is needed:

* one predicate decides both the route and the layout. Four inline copies of that condition, their four
  now-dead per-file `kTmaBlockM` constants, and a fifth copy in the fused SwiGLU route are gone;
* the tile geometry is named once in `nvfp4_config.h`, beside the family's other layout constants, and
  both descriptor builders and both load sites use it;
* the layout is a named `Nvfp4ScaleLayout`, on the host and in the kernel, so the two routes that force
  one say which;
* both descriptor builders `static_assert` their `BlockM` against the tile width the quantizer writes.
  They previously admitted 128, which no registered schedule uses and which would have addressed those
  tiles mis-shaped;
* the tiled layout is a bijection onto the plane only for a whole number of token tiles — at
  `tokens = 257, K = 5120` the last tile's write lands at byte 159 759 of an 82 240-byte plane — so the
  launcher checks that where the layout is acted on, rather than trusting each caller's own predicate.

Deleting the per-file constants re-aligns the `using` block above each; that is `clang-format`, not a
change, and it accounts for about 80 of the diff's lines.

## Result

**Operator.** `ninfer_linear_bench --qtype NVFP4 --policy a4 --execution graph --warmup 5 --repeat 30
--flush-mib 256`, every cell its own process, seven passes with the arm order rotated. A third arm is
the master binary under a second label as the identical-baseline null.

| shape | T = 1024…8192 | range | cells faster |
|---|---:|---|---:|
| `14336 x 5120` | **−8.2 %** | −9.0 … −6.3 | 4/4 |
| `16384 x 5120` | **−4.9 %** | −8.8 … −1.9 | 4/4 |
| `34816 x 5120` | **−7.7 %** | −8.2 … −7.0 | 4/4 |
| `5120 x 6144` | **−8.4 %** | −8.9 … −7.9 | 4/4 |
| `5120 x 17408` | **−11.1 %** | −13.6 … −9.2 | 4/4 |

All 20 route-taking cells are faster, median **−8.2 %**. **T = 768 is below the route floor and reads
0.00 %** (−0.15 … +0.07), which is what shows the two builds are otherwise the same.

**Kernel time in the model.** Qwen3.8-27B NVFP4, `-p 16384 --prefill-chunk 4096 --kv-dtype int8
--max-ctx 20480`, `nsys`, two passes with the order mirrored. The NVFP4 kernels are 25.7 % of all
kernel time:

| | master | this change | null |
|---|---:|---:|---:|
| NVFP4 kernels, pass 1 | 1390.99 ms | **−3.37 %** | −0.20 % |
| NVFP4 kernels, pass 2 | 1378.56 ms | **−3.87 %** | +0.60 % |
| all kernels, pass 1 / 2 | 5403.74 / 5364.23 ms | −0.94 % / −0.80 % | −0.25 % / +0.40 % |

Both sides of the trade, same runs: the two TMA GEMM kernels go 1322.1 → 1274.8 ms (**−3.58 %**,
−4.12 % on the second pass) and the quantizer that writes the plane goes 63.1 → 63.8 ms (**+1.06 %**,
same 1344 launches). Tile-contiguous, a warp's 32 scale bytes are two 16-byte runs 4080 bytes apart
instead of one contiguous store; that cost is real and it is inside every number here. It is 0.6 ms
against 47 ms.

**End to end.** Same artifact, `-p 32768 --max-ctx 36864 -r 5 --warmup 2`, three passes with the order
rotated, `prefill_tok_s_mean`:

| prefill chunk | this change | null |
|---|---:|---:|
| **1024 (default)** | **+1.01 %** (+0.04, +1.32, +1.01) | −0.23 % |
| 4096 | **+0.82 %** (+0.51, +0.93, +0.82) | −0.12 % |

Six readings of six are positive, but at chunk 4096 the worst of them (+0.51 %) only just clears the
best null reading (+0.47 %), and one null pass at chunk 1024 went to −1.90 %. **Separation is not full
and I am not claiming it**; the end-to-end figures confirm the direction, and the kernel-time table is
what carries the claim.

The model gains less than the fixture, and the reason is in the fixture: `--flush-mib 256` evicts the
scale plane before every call, while in the model the quantizer has just written it and it is still
resident — so the fetch this change improves is already cheaper there.

`ee9d5192`, merged since this work began, reduces **how often** the scale box is fetched; this changes
**the shape of each fetch**. The two compose, and what is measured above is what is left after it. The
roughly twice larger figures from before that commit are not repeated.

## Validation

Bitwise identical, and structurally so: values do not change, only the address a scale byte is written
to.

The gate is the five NVFP4 Op tests run with `NINFER_OP_REPORT_STATS=1`, which prints `max_abs`,
`max_rel` and `rel_l2` per case at `%.17g`, compared byte for byte: **3007 numeric records identical**
on both arms (15 / 24 / 26 / 2816 / 126), all five reporting OK, with the five pairs of binaries
checked to differ first.

The gate is shown able to fail. A build with the tile mapping damaged — pairs of groups swapped inside
a tile, which stays in bounds — moves **exactly 17 of the 3007 records, and all 17 are route-taking
cases**: T = 1024 in all five tests plus T = 256 in the fused SwiGLU, whose own floor is 256. The other
2990 do not move, because at those widths the route is not taken. **Two of the five tests still
reported OK under that damage** — their tolerances absorbed it — so the record comparison, not the pass
token, is what carries the claim.

`ctest` is 115/115 on both arms from the same build flags, with the same seven artifact-gated tests skipped on both. `clang-format` clean. Resource deltas are zero: workspace capacity and peak,
weight capacity and KV payload are byte-identical as the engine reports them, and no kernel changes
registers or shared memory. The binary gains three kernel instantiations — the quantizer template takes
the layout, so each of the three activation geometries is emitted twice, at 40 registers and no shared
memory.

## Limits

* The route is unchanged: this changes how it is fed, not which widths it admits. Below the floor of
  1024, and at lengths that are not multiples of 256, nothing moves — which is most real prompt
  lengths. The follow-ups below are about that.
* No roofline share is quoted. The NVFP4 tensor-core ceiling I have is from a separate probe on this
  part, not from this session, and the bench's own ceilings cover DRAM and dense BF16 only.
* The card is shared with a desktop compositor, which preempts a run outright rather than slowing it.
  That is what the identical-baseline arm is for: it dropped 45 of the 175 operator passes, and the
  null of the survivors has a p95 of 0.79 %. SM clock over the sweep was 2430–2760 MHz.

## What follows on this route

Three further changes to the same dispatch and descriptor, each to be measured and submitted separately
on the base of the day. They are sequenced behind this one because they move the same code, and each is
worth more once the feed is fixed. The ranges below were measured on `ad0f3d38` and will be
re-established at submission:

1. **Admit a partial last M tile** — drop `tokens % 256 == 0`. The largest of the three on real traffic,
   since prompt lengths are rarely multiples of 256: +16 to +26 % prefill at chunk 8192.
2. **Lower the route floor** from 1024 to 640: +2.7 to +7.2 % in the 640–1023 band, flat elsewhere.
3. **Correct three schedule thresholds below 400 tokens**: +3.0 to +18.6 % at 56–96 and 224–352.

Unlike this PR, all three change which kernel runs, so each is qualified against the Op's own oracle
rather than by byte comparison.

RTX 5090 `sm_120a`, CUDA 13.1, Release, base `d4929686`.


MichaelDementii:
Related: #239 changes the activation in the epilogue of the fused SwiGLU kernel this PR also touches,
in a different place in the same file. They merge cleanly in either order; whichever lands first
leaves the other a trivial rebase.

MichaelDementii:
This is one of four changes to the same NVFP4 W4A4 route, numbered in the order they are meant to be
read. They are separate PRs because they are separate mechanisms, but the order is not arbitrary:

   shares a file with #238.
3. **#240** — the route stops needing a token count that is a multiple of 256. Depends on #238:
   that PR put the route predicate in one place, so the widening is one line rather than four, and
   its tiled scale layout is what #240 extends to a partial tile.
4. The route floor, still to be re-measured and submitted. It is worth little before #240 — below
   1024 only one width is a multiple of 256 — and the whole band after it.

Any of them can be reviewed alone; #240 carries #238's commit until that one lands.

MichaelDementii:
Closing in favour of #248, and opening #247 for the scope, as the template now asks.

Master's NVFP4 reorganization moved this change's ground: `nvfp4_config.h` was emptied, the per-shape
`shapes/n*.cu` files took over route selection, and `perf(ops): tune nvfp4 34816x5120 linear routes`
lowered that shape's TMA floor to 256 while the other four stayed at 1024. A rebase resolved into
re-adding structures you had just deleted, so I rewrote the change against the new layout instead of
patching it through.

The rewrite is also better than what was here. The scale layout and the route must agree — a
row-major plane read as tiled is wrong numbers, not a failure — and this version makes them
inseparable: `select_a4` returns the GEMM together with the layout it reads, so a shape cannot pair
them the other way round. Everything was re-measured on `e360c4c0`, the quantizer is now timed
separately from the GEMM so the write side's cost is stated rather than netted away, and the route
boundaries are covered in all five Op families that share the plane — which turned up a pre-existing
sizing bug in the `linear_add` test harness, fixed there.

Nothing here is withdrawn as wrong; it no longer applies to the tree.

PR #240, opened 2026-09-14
NVFP4 W4A4 route 3/4: take a partial last M tile, so the route stops needing a multiple of 256

MichaelDementii:
Third of the NVFP4 W4A4 route series, after #238 and #239. **Depends on #238**, whose commit is the
first of the two here; when that merges this rebases to one commit.

Why it is sequenced that way rather than independent: #238 put the route predicate in one place, so
the widening below is one line instead of four, and #238's tiled scale layout is what this change has
to extend to a partial tile.

## Problem

The route requires `tokens % 256 == 0`. Prompt lengths are rarely multiples of 256, and a prefill
chunk only has to be a multiple of 128, so on most real widths the fast route is simply not taken —
including every chunk of a serving configuration whose chunk size is not a multiple of 256.

## Change

partial. Three things follow from that, and they are the substance of the diff:

* **The scale plane is padded to whole tiles.** #238's tiled layout is a bijection onto whole tiles
  only, so the plane is allocated over `nvfp4_w4a4_padded_tokens` — at most 255 tokens of scales,
  under 0.3 MiB on the widest registered K — and the quantizer writes zeroes into the padding. The
  consumer's box covers the whole last tile and would otherwise read what the arena last held. The
  padded rows own no input and no code byte; their codes come back as TMA zero-fill, and a zero scale
  times any E2M1 code is exactly zero, so a padded row contributes nothing.
* **The workspace planners scan instead of rounding.** Both the fused and the baseline SwiGLU
  capacity used to round a whole-tile width down or subtract one, which was correct only while the
  route was confined to multiples of 256. The baseline one in particular would otherwise reserve
  `baseline_workspace_bytes(max_tokens - 1)` for a width that can no longer reach the baseline route
  — about 540 MiB at chunk 8192, out of the same budget as the KV cache.
* **The fused SwiGLU route gets its own rule**, one whole tile or at least one and a half, because
  between 257 and 383 a padded tile is a third of the work.

The shared route takes no floor beyond the existing 1024: measured below, the emptiest admitted tile
is still faster.

## Result

Operator against #238 alone, `ninfer_linear_bench --qtype NVFP4 --policy a4 --execution graph
--warmup 5 --repeat 30 --flush-mib 256`, five registered geometries, seven passes with the arm order
rotated, third arm the #238 binary again as the identical-baseline null.

| T | 1153 | 1408 | 1665 | 2049 | 3073 | 4097 |
|---|---:|---:|---:|---:|---:|---:|
| ragged | **−17.7 %** | **−19.7 %** | **−24.4 %** | **−21.3 %** | **−21.2 %** | **−23.2 %** |

30 ragged cells, 29 faster, median **−21.8 %**. The aligned widths are the control and do not move:
T=1024 **+0.05 %**, T=2048 **−0.01 %**, T=4096 **+0.00 %**, 15 cells, median +0.00 %.

The widths where the padding is the largest share of the work are the ones to worry about, so they
were swept separately:

| T | padding | |
|---|---|---:|
| **1025** | 255 of 1280, 19.9 % | **−10.6 %** |
| 1281 | 255 of 1536, 16.6 % | −18.8 % |
| 1537 | 255 of 1792, 14.2 % | −21.6 % |
| 2305 | 255 of 2560, 10.0 % | −20.5 % |
| 4353 | 255 of 4608, 5.5 % | −23.2 % |

25 of 25 faster, median −19.2 %. The gain grows as the padding share falls, which is the shape the
mechanism predicts, and even the emptiest tile is well clear of the null (p95 0.79 %).

**End to end**, Qwen3.8-27B NVFP4, `--kv-dtype int8 --max-ctx 20480 -r 5 --warmup 2`, three passes
with the arm order rotated. `--prefill-chunk` must be a multiple of 128 but not of 256, so 3968 makes
every chunk ragged and 4096 is the aligned control:

| prefill chunk | prefill throughput | null |
|---|---:|---:|
| **3968, every chunk ragged** | **+12.7 %** (+12.27, +14.44, +12.73) | +0.30 % |
| 4096, control | **+0.03 %** (−0.26, +0.03, +0.13) | +0.02 % |

## Validation

This changes which kernel runs, so the qualification is the Op against its own FP64 oracle, not a
byte comparison.

**The ragged path was not executed by any test before this change** — the only route-taking width in
the NVFP4 suite was 1024, a whole number of tiles. Added: **T=1025** on all four linear geometries,
on `attn_input_proj`, on `gdn_input_proj` and on `linear_add`, and 400 and 1025 on the fused SwiGLU.
1025 is both the first width the route admits with a partial tile and the emptiest — one real token
in the last tile and 255 of padding. `linear_add` is the case that most needed it: it is the only
in-place route, and the only epilogue that reads the residual at the token index the kernel clamps.

All five report OK against their oracles. `ctest` is 115/115. `clang-format` clean, no line
over the column limit added.

**The new width is shown to catch what the old ones do not.** With the scale descriptor's extent
taken from the real token count instead of the padded one — so the last partial tile falls outside
the declared region — the suite fails at T=1025 on all five geometries and nowhere else:

```
NVFP4_A4 [14336,5120] T=1025: reduction criterion failed  actual=0  reference=-41.1
NVFP4_A4 [16384,5120] T=1025: ...                         actual=0  reference=-53.9
NVFP4_A4 [34816,5120] T=1025: ...                         actual=0  reference=-50.1
NVFP4_A4 [5120,6144]  T=1025: ...                         actual=0  reference=-77.1
NVFP4_A4 [5120,17408] T=1025: ...                         actual=0  reference=+113.9
```

One fix to the harness came out of adding the width: `tests/ops/linear_add/test_nvfp4.cpp` sized its
host activation and residual from a literal 1024 while the invocation list is what actually drives
them, so any width past that read past the end of the buffer. The bound is now taken from the list.

## Limits

* The padding is not free: at T=1025 four fifths of the last tile is padding and the gain is
  −10.6 % rather than the −21 % the fuller widths see. It is still a gain at every width measured.
* The scale plane grows by up to 255 tokens of scales, under 0.3 MiB on the widest K, and it is paid
  whichever layout the quantizer then writes.
* Measured on one card whose desktop compositor can preempt a run outright; that is what the
  identical-baseline arm is for. It dropped 64 of 315 operator passes on the first sweep and 35 of
  175 on the second, and the null of the survivors has a p95 of 0.79 %.
* The end-to-end figure is on one artifact and one prompt length.

## What follows

One more on this route, kept separate because it is a threshold and not a mechanism: the floor of
1024 was measured at 640 on an older base. Before this change, lowering it would have admitted a
single new width, 768; after it, the whole band. It will be re-measured on the base of the day and
submitted on its own.

RTX 5090 `sm_120a`, CUDA 13.1, Release, base `d4929686` plus #238.


MichaelDementii:
This is one of four changes to the same NVFP4 W4A4 route, numbered in the order they are meant to be
read. They are separate PRs because they are separate mechanisms, but the order is not arbitrary:

   shares a file with #238.
3. **#240** — the route stops needing a token count that is a multiple of 256. Depends on #238:
   that PR put the route predicate in one place, so the widening is one line rather than four, and
   its tiled scale layout is what #240 extends to a partial tile.
4. The route floor, still to be re-measured and submitted. It is worth little before #240 — below
   1024 only one width is a multiple of 256 — and the whole band after it.

Any of them can be reviewed alone; #240 carries #238's commit until that one lands.

MichaelDementii:
The fourth and last change on this route is now open as #241: the floor of 1024, re-measured on this
base as promised above. The old number this package carried, 640, does not survive the re-measurement
— it regresses `gdn_input_proj` — so the floor proposed there is 768, the first whole number of tiles
above the last measured regression.

MichaelDementii:
Closing in favour of #253, with the scope in #252.

Master's NVFP4 reorganization moved the ground under this one: the route condition it edits now lives
in each shape's `select_a4`, and `perf(ops): tune nvfp4 34816x5120 linear routes` put that shape's
floor at 256 — which is exactly where a ragged width hurts most, because it falls the whole way back
to the MMA route. Rebasing produced conflicts against structures you had deleted, so it was rewritten
against the new layout on top of #248 rather than patched through.

Re-measured on `e360c4c0`: fifteen ragged widths of fifteen are faster, median −17.5 %, best −48.1 %
on `34816x5120`, and the fifteen whole-tile widths take the same route on both arms and read 0.00 %
as controls.

Two things the new body says that this one could not. The numeric witness now compares both arms on
identical test widths and reports that exactly two records of 3293 differ — the two fused-SwiGLU
widths whose route actually changes, where the fused path's error is the smaller of the two — rather
than claiming bit-identity everywhere. And the fused route's workspace capacities are found by asking
the route across the range instead of by rounding, which is what the review of this PR caught as a
540 MiB regression the first time.

PR #241, opened 2026-09-14
NVFP4 W4A4 route 4/4: lower the route floor to three whole M tiles

MichaelDementii:
Last of the NVFP4 W4A4 route series, after #238, #239 and #240. **Depends on #240**: this branch
carries that PR's two commits until they land, and this change is the third.

## Problem

The shared W4A4 TMA route is offered only from 1024 tokens. That floor predates #240, which let the
route take a partial last M tile, and no measurement supports it.

## Change

One constant: the floor becomes `3 * kNvfp4TmaBlockM`, and the route predicate reads it instead of a
literal 1024.

## Result

At 768 tokens `linear` is 14.5 to 43.7 % faster across five geometries, median 22.6 %, and the three
other Ops that share the floor are faster by a median of 16.9 %, worst cell 13.6 %, on their own
benchmarks. End to end, a prefill chunk of 768 on Qwen3.8-27B NVFP4 is 1.9 % faster.

All figures below are medians over the passes stated, arm order rotated, with the base binary run
again under a second label as an identical-baseline null; passes whose null exceeds 1 % are dropped.

## Why 768 and not lower

`ninfer_linear_bench --qtype NVFP4 --policy a4 --execution graph --warmup 5 --repeat 30 --flush-mib
256`, five registered geometries. **The route does not improve monotonically in T.** Over 256..1023,
four passes, every width has a negative median — and seven cells still regress:

| T | geometry | Δ | surviving passes |
|--:|---|--:|--:|
| 320 | `14336x5120` | **+35.1 %** | 2 |
| 384 | `14336x5120` | **+34.5 %** | 3 |
| 448 | `5120x6144` | **+19.5 %** | 4 |
| 448 | `5120x17408` | **+18.8 %** | 3 |
| 512 | `5120x6144` | **+23.7 %** | 3 |
| 512 | `5120x17408` | **+18.9 %** | 4 |
| 640 | `16384x5120` | **+2.7 %** | 4 |

A floor chosen on the median would ship all seven. The two geometries that lose at 448 and 512 win
again at 640: the sign is not monotone, so a width that measures well does not vouch for its
neighbours.

At the boundary, seven passes per cell:

| T | 5120×6144 | 5120×17408 | 14336×5120 | 16384×5120 | 34816×5120 |
|--:|--:|--:|--:|--:|--:|
| **640** | −20.8 | −25.1 | −20.1 | **+3.5** | −32.9 |
| 672 | −19.7 | −25.8 | −21.2 | −13.1 | −44.3 |
| 704 | −20.5 | −25.6 | −21.4 | −14.2 | −44.2 |
| 736 | −17.9 | −24.8 | −21.2 | −13.7 | −43.9 |
| **768** | −20.2 | −25.4 | −22.6 | −14.5 | −43.7 |
| 832 | −16.9 | −22.9 | −14.3 | −6.4 | −36.5 |

640 is the last width with a regression, and 672, 704 and 736 are faster everywhere — **so 768 is not
the largest admissible floor.** It is the first whole number of tiles above the last measured
regression, taken deliberately. #200's post-mortem showed a crossover of this kind moving with SM
clock; with one machine I cannot establish the binding case, so the band 672..767 is given up
knowingly rather than claimed on a single card's reading.

**The floor is shared by four call sites, and the sweep above measured one.** The others differ in
output policy — `attn_input_proj` scatters one row block into four destinations, `gdn_input_proj`
into two, `linear_add` reads the residual per token in place. Their own benchmarks, five passes, four
bench cells over the three Ops:

| T | median | worst cell |
|--:|--:|---|
| **640** | −14.7 % | **`gdn_input_proj` +1.8 %** |
| 672 | −16.5 % | −11.6 % |
| 704 | −15.3 % | −11.5 % |
| **768** | **−16.9 %** | **−13.6 %** |
| 832 | −14.1 % | −6.2 % |

The regression at 640 therefore reproduces on a second instrument: the `linear` sweep found it on the
16384×5120 geometry, and `gdn_input_proj` — which is that shape — finds it in its own benchmark. One
shape, two kernels and two output policies, the same sign. That is also why the floor is not 672.

**Controls.** Above the floor both arms run the same kernel by construction. `linear` reads +0.00 %
at 1025 and −0.01 % at 2048; the other three read 0.00 % across all four bench cells at 1024, and
+0.01 / +0.07 / 0.00 / −0.02 % at 2048.

**Discard rates**, since the gate is aggressive: 36 of 200 in-band passes dropped (null p95 0.83 %),
63 of 210 at the boundary (p95 0.56 %), 28 of 140 on the other Ops (p95 0.73 %). One cell lost every
pass — 768 on `5120x17408` in the four-pass band sweep — which is why that width was re-run at seven
passes; it reads −25.4 % there.

## End to end

Qwen3.8-27B NVFP4, 15360-token prompt, `--kv-dtype int8 --max-ctx 20480 -r 5 --warmup 2`, six passes:

| prefill chunk | prefill throughput | null |
|---|---:|---:|
| **768** | **+1.9 %** (+1.80, +4.02, +1.91, +1.28, +1.71, +2.73) | −1.01 % |
| 1024, control | +0.04 % | −0.39 % |

Six of six positive and the worst, +1.28 %, clears the best null reading, +0.23 %. It is not a tight
measurement — the null for that chunk ran −1.45 to +0.23 % and is not centred on zero — so roughly a
percent of the figure is instrument. The claim rests on the operator tables; this shows it survives a
whole model.

## Validation

The floor is a performance threshold, not a correctness one: both routes are already qualified
against the same FP64 oracle, so no test can fail on the value itself. What the tests must do is
cover the configuration the new floor newly reaches.

`tests/ops/linear/test_nvfp4_a4.cpp` gains 767 and 769 on all four call sites, and
`test_attn_input_proj.cpp` gains 769. **769 is the emptiest grid the route will ever run**: three
full tiles and a fourth holding one token, so 255 of its 1024 tile slots are padding, against 255 of
1280 at 1025 — the previous worst case. #240 made those padded rows contribute exactly zero through a
zero scale; this is the harshest test of that invariant the route admits. 767 keeps the replaced
route covered immediately below the floor.

A single case at 768 would not have done: it is three whole tiles and configuration-identical to the
1024 case already present.

`ctest` 115/115, the same 7 artifact tests skipped as on the base. `clang-format` clean, no
added line over the column limit.

## Limits

* **The floor is conservative by 96 tokens** above the smallest value this sweep supports, and by
  half a tile above the last measured regression.
* One machine, and the sign is not monotone below 640, so this is evidence about these widths on this
  card rather than a model of the crossing. Two of the regressing cells rest on two or three
  surviving passes; their magnitudes are not to be relied on, only their sign, which is far outside
  the null.
* Which real widths land in the opened band: a configured prefill chunk of 768, and the final ragged
  chunk of any prefill whose remainder falls in 768..1023. A power-of-two chunking policy reaches
  only the first.
* The fused SwiGLU route keeps its own, lower floor and is untouched: its non-TMA widths, 129..255
  and 257..383, lie below both.
* The end-to-end figure is one artifact, one prompt length, one chunk width in the band.

RTX 5090 `sm_120a`, CUDA 13.1, Release. Base `d4929686`.


MichaelDementii:
This is the last of four changes to the same NVFP4 W4A4 route, numbered in the order they are meant
to be read. They are separate PRs because they are separate mechanisms:

   shares a file with #238.
3. **#240** — the route stops needing a token count that is a multiple of 256. Depends on #238.
4. **this one** — the route floor, 1024 → 768. Depends on #240: before it, the band this opens held
   exactly one admissible width, so the two are worth little apart.

Any of them can be reviewed alone; this branch carries #240's two commits until that lands, and #240
carries #238's.

@codex review

MichaelDementii:
Closing this one without resubmitting it.

The proposal here was a single floor for the shared W4A4 TMA route, measured when all five registered
shapes used the same 1024. That is no longer the shape of the question. After
`refactor(ops): organize nvfp4 execution by matrix shape` the floor lives in each shape's own
`select_a4`, and `perf(ops): tune nvfp4 34816x5120 linear routes` has already put that shape at 256
while the other four stayed at 1024. A global threshold is the wrong unit now, and re-measuring one
per shape would be duplicating the tuning you are doing yourself, with your own priorities, anchors
and report format.

What was actually load-bearing in this branch was not the number. It was that the route cannot go
below a multiple of 256 at all until it can take a partial last M tile — which is the separate change
this one was stacked on. I am preparing that one against the new base; with it, the per-shape floors
you are already tuning become free to move to any width rather than to the next multiple of 256.

The measurements here are not withdrawn as wrong, and they are recorded in the thread if the crossover
is ever useful: the last width at which the replaced route still won was 640, and 672 through 768 were
faster on every registered geometry on this card. They simply describe a tree that no longer exists.

PR #242, opened 2026-09-14
sparse_moe: widen the Q5 routed-down Rows2 window 5 to 17, its measured crossover

MichaelDementii:
`resolve_sparse_moe_small_t_plan` picks the D4 row split from a per-codec token threshold. For
`Q5G64_F16S` that threshold is 5. The measured `Rows2`/`Rows4` crossover on that codec is much
further right, so widths 6..17 run the slower schedule.

## Change

One number: the `Q5G64_F16S` window goes **5 → 17**. `Q6G64_F16S` keeps the 11 it has. `Rows` is only
a grid split — `(kHidden / Rows, tokens)` — and both instantiations already exist for both codecs.

The threshold is per codec because `routed_down_format` on Qwen3.6-35B-A3B picks Q6 for layers 34, 38
and 39 and Q5 for the other 37, so both branches run inside the same decode round.

## Result

Q5, concentrated routing — every token sharing one set of eight experts, the pattern where `Rows2` is
weakest — on the slower of two cards, median of the kept per-pass ratios:

| T | 6 | 8 | 10 | 12 | 14 | 16 | 17 | 18 |
|---|--:|--:|--:|--:|--:|--:|--:|--:|
| Δ | **−2.63 %** | **−2.43 %** | **−1.82 %** | 0.00 % | −0.02 % | 0.00 % | 0.00 % | **+1.09 %** |

So the gain is 1.8–2.6 % at T=6..10, and from 12 to 17 the two schedules are **indistinguishable at
this instrument's resolution**: the readings sit on a 2.048 µs lattice and the operation is 133 µs at
T=17, so one step is 1.18 % and nothing smaller can be resolved. T=18 is the first loss, and it is
outside the window.

On trace-like routing the same range reads −1.3 % to −8.8 %. Concentrated routing is the worst case
and sets the threshold; trace-like is where the gain actually accrues.

Largest T at which `Rows2` is not slower, by routing distribution, T ≤ 2 excluded (that is `Rows1`),
threshold 0.50 %:

| codec | trace-like | independent | **concentrated** |
|---|---:|---:|---:|
| Q5, 1975 MHz | ≥22 | 18 | **17** |
| Q5, 2810 MHz | ≥22 | 20 | 19 |
| Q6, 1975 MHz | 15 | 16 | **11** |
| Q6, 2810 MHz | 17 | 22 | 16 |

Concentrated binds everywhere, and the lower-clocked card binds over the higher: the cost of `Rows2`
is SM-side, so at a lower clock it is larger in absolute time and the crossover moves left. `≥22` is
the edge of the sweep, not a measured crossover.

**Against the machine**, measured on this base at concentrated routing: T=6, 12 and 17 read 328.1,
191.4 and 141.3 GB/s of unique weight — 18.3 %, 10.7 % and 7.9 % of the 1792.1 GB/s ceiling the bench
reports — at 6.03, 7.04 and 7.37 logical TFLOP/s. Nothing here is bandwidth-bound, which is why a
grid split moves it at all.

## Why Q6 is not moved

An earlier revision of this change raised Q6 as well, and review found a 1.61 % operator regression at
Q6/concentrated/T=17. That reproduced here: +1.56 % on a 2810 MHz card and +2.33 % on a 1975 MHz one,
identical-binary control 0.00 % in all three.

A cutoff of 16 was suggested and does not hold either. On the 1975 MHz card Q6 regresses at **T=12
(+1.56 %)** and **T=16 (+2.45 %)**; T=14 reads +0.04 %, inside the null band; T=15 was not measured.
The largest value that regresses on neither card is 11, which master already has, so Q6 is left alone
and only Q5 moves.

**That has a price, and it is not zero.** On the 2810 MHz card Q6's own crossover is 16, so on faster
cards three of the forty layers keep `Rows4` where `Rows2` would serve them. Leaving Q6 at 11 is the
conservative choice, not the optimal one.

## Validation

Bit-exact by construction: `Rows` changes only how output rows are distributed over blocks. Each row
accumulates over the intermediate dimension in the same order, and the nine paths are summed in rank
order in the epilogue, independent of `Rows`.

Witnessed where the change actually acts. `--spec dflash --draft-tokens 11` and `15` put decode at
T=12 and T=16, inside the moved window, so master runs `Rows4` and this branch runs `Rows2` on every
decode step. Three prompts × two widths, greedy, 35B artifact: **6 of 6 completions byte-identical**,
448–604 bytes each.

That gate is shown able to fail. A third arm with the `Rows2` epilogue scaled by 1.001 — the schedule
this change selects, and nothing else — **differs from master in 4 of the same 6 cells**. Two cells
absorb a perturbation that small, so the witness is not sensitive to everything; it is sensitive to
this. It also establishes that the branch really executes `Rows2` at those widths, since a `Rows2`-only
perturbation could not otherwise change the output.

`ctest` 115/115. `clang-format` clean, columns verified independently, no added line over 100.

## Limits

* **The arms are separate binaries** from one tree with a byte-identical rebuild as the null control,
  not all routes in one process. The in-process protocol was not used here. On the cell that started
  this, the two methods agreed closely in review — +1.6125 % in process against +1.613 / +1.613 /
  +1.587 % from separate binaries.
* The binding card discarded **164 of 846 passes** on its null control (19.4 %); the faster card
  discarded 7 of 391.
* Readings below 1.18 % of the operation are one lattice step and cannot be resolved, so the zeros in
  the T=12..17 row mean "not distinguishable", not "proven equal".
* Per-cell coverage exists for Q5/concentrated on the binding card, the row above. The other five
  codec × distribution × card combinations are summarised by their crossover only, and a crossover
  does not by itself exclude a slower cell below it.
* Every width in [2,46] is reachable: the small-T route is selected by token count alone, and the tail
  chunk of a prefill enters it as well as decode at batch ≥ 2.
* **No end-to-end number.** At batch 1 the decode round runs T=4, inside the old window as well as the
  new one; T=6..17 needs batch ≥ 2 or speculation, and this card cannot resolve a per-layer operator
  effect of this size end to end.
* Measurements were taken on `b88c0f6f`. Master is `d4929686`; the two commits between them touch no
  file under `src/ops/sparse_moe/`, and the branch is rebased onto it. The roofline and witness rows
  above were taken on the rebased branch.

RTX 5090 `sm_120a`, CUDA 13.1, Release, `qwen3_6_35b_a3b.ninfer`, `ninfer_sparse_moe_bench`.

Generated with Claude Opus 5.


MichaelDementii:
Supersedes #200, which carried a body written for two earlier revisions of this change. The review history and the measurements in it stay there.

@codex review

MichaelDementii:
Closing in favour of #255, with the scope in #254.

The change itself survives master's rework — the only conflict was the v3 rename of
`QType::Q5G64_F16S` to `Q5_G64_FP16` — but every number here was measured on `d4929686`, so none of
them belongs to the tree any more.

Re-measured on `e360c4c0`, and the re-measurement changed what the body can claim. This one sampled
the range through every other width; the new one measures **every integer from 3 to 20**, which is
what the threshold actually needs, and it shows that the earlier picture was an artifact of the
spacing: 16 and 17 gain 1.5–1.6 % while 12, 14 and 15 read zero. The zeros are the instrument — every
absolute median is an exact multiple of 2.048 µs, so at T=12 one step is 2.08 % of the operation and
anything smaller cannot be resolved. The claim is now per cell: no width in 6..17 regresses.

Also new: the Q6 branch, which this change does not touch, is swept as a control and reads 0.00 % at
all eighteen widths, and the output witness reports the byte size of what it compared, because two
empty captures compare equal just as happily as two matching ones.

PR #243, opened 2026-09-14
W8 GEMM: choose the predicated cache policy instead of inheriting it, operator -3.3 to -3.7% on ragged token counts

MichaelDementii:
In `w8_rowsplit_gemm_mma_kernel` the full-tile branch spells `Cache::cg` and the predicated branch a
few lines below it does not spell anything — it inherits `Cache::ca` from `cp_async_zfill`'s default.
Which branch runs is decided per launch, not per tile, so any token count that is not a multiple of
`BN` sends the whole GEMM down the predicated path with a policy nobody chose.

## Change

`kPredicatedCache` becomes a schedule parameter **defaulting to `Cache::ca`, so adding it changes no
existing instantiation**, and exactly one schedule opts in: `MmaR64C128`, the wide route of the linear
family. The opt-in goes through a member alias template so the trailing parameters, and their
defaults, cannot be respelled at the point of use. Two files, +32/−6.

Six other call sites instantiate the same `<64, 128, 64, 16, 2, 2>` geometry — `linear_add`,
`gdn_input_proj`, `attn_input_proj` (twice), `dflash2_attn_input` and `linear_pair`. They do not opt
in, so they compile to exactly what master has. Whether they would also gain is unmeasured and is a
separate change.

## Result

`ninfer_linear_bench --qtype W8 --execution graph --warmup 20 --repeat 300 --flush-mib 256`, three
passes with the arm order rotated, ratios paired inside a pass, median over passes then over T; the
null is a second run of the master binary under a different label.

| shape | T | grid, CTAs | this patch | null p95 |
|---|---|---|---:|---:|
| feature 5120×25600 | 129..255 | 80..160 | −11.47 % | 0.045 % |
| feature 5120×25600 | 257..383 | 240 | −1.82 % | 0.013 % |
| generic 9216×2048 | 4096..4223 | 4752 | **−3.35 %** | 0.961 % |
| mid 34816×5120 | 129..319 | 1088..1632 | **−3.73 %** | 0.365 % |
| lm head 248320×5120 | 129..319 | 7760..11640 | **−3.74 %** | 0.468 % |
| sparse 2048×4096 | 896..1096 | 224..288 | 0.00 % | 0.026 % |

**The −11.47 % is not the headline.** The feature shape runs 80 to 160 CTAs on a 170-SM part — it does
not fill the machine once — and the gain falls to −1.82 % as soon as the grid reaches 240. The number
to carry is **−3.3 to −3.7 %**, on the three shapes whose grids run to thousands of CTAs.

The sparse shape does not move at all. It reaches the same schedule and reads 0.00 % against a
0.026 % null, so the gain is not a property of the route alone.

**Control.** `full` requires `(T % BN) == 0`, so T=4096 is the one point in the generic sweep the
patch cannot act on. It reads −0.26 % — inside that shape's own 0.96 % null, which is the widest in
the table — while the swept range around it reads −3.35 %. The generic range covers all 128 `BN=128`
alignment phases once each.

**Against the machine**, from the same rows: 8.0 % to 23.7 % of the 1689.4 GB/s read ceiling, so
nothing here is bandwidth-bound, and 29 % to 84 % of the dense bf16 tensor peak. That peak is
236.7 TFLOP/s at the 2720 MHz mean SM clock recorded during this run, not the 209.5 of the nameplate
clock. The generic shape sits at **84 % of peak on master and still gains 3.35 %** — this is not a
roofline wall moving, it is L1 tag and allocation work removed from a path that gets nothing from L1.

## Validation

A cache policy selects where a load is cached, not what it returns, so the output must be bit-identical
— witnessed rather than argued. The six Op tests covering the translation units that include this
header, run on both arms with `NINFER_OP_REPORT_STATS=1` at `%.17g`: **4145 numeric records,
byte-identical** (600/123/288/192/2816/126). `ctest` 115/115 on both arms. `clang-format` clean,
columns verified independently.

"Changes no instantiation but the one that opts in" is checked in the object rather than asserted.
`cuobjdump -sass` on both builds: **30 function bodies and 454 `LDGSTS` on each side, and exactly one
body differs** — the predicated BM=64 BN=128 one, whose 15 `LDGSTS` go from 0 to 15 `BYPASS`. Every
mangled name changes, because the schedule type gained a parameter, so the comparison is by body
content with addresses and encodings stripped.

## Limits

* **No end-to-end claim.** This is an operator-level change; the shapes that move are not reachable by
  an end-to-end run on the artifacts on this machine.
* **The arms are separate binaries** built from one tree, with a second run of the master binary as
  the identical-baseline null, rather than all routes captured in one process. The policy is a
  compile-time template parameter, so one binary would need a runtime switch that does not exist.
* The card discarded **137 of 543 passes** on the null gate (25 %). The generic shape's surviving null
  has a p95 of 0.961 %, wide enough that its −3.35 % is about 3.5 nulls, not more.
* Five of the nineteen dispatch cells reaching `MmaR64C128` are measured, covering five of its fifteen
  distinct (n, k) shapes.
* Every other predicated instantiation keeps `ca`, **including ones separately measured to gain from
  `cg`**. Moving them is a separate change with its own per-shape measurement.
* ncu wall clock is not quoted anywhere: it serialises launches and controls caches between replays,
  which removes the L2 reuse at issue.

RTX 5090 `sm_120a`, CUDA 13.1, Release, one card, one session, SM clock 2475–2842 MHz, mean 2720.
Base `d4929686`.

Generated with Claude Opus 5.


MichaelDementii:
Supersedes #201, which carried a body written for an earlier revision whose default was `cg`. The review history and the measurements in it stay there.

@codex review

MichaelDementii:
Closing in favour of #257, with the scope in #256.

Master renamed the whole family `W8` → `Q8` and moved the directory, so the files this PR edits no
longer exist under those names and a rebase resolved into "deleted by us / modified by them". It was
rewritten against `q8_rowsplit_gemm_mma.{cuh,cu}` rather than patched through. The inconsistency it
addresses survived the rename word for word: the predicated path still inherits `Cache::ca` while the
full path two lines above spells `cg`, and seven call sites still instantiate the same tile.

Everything was re-measured on `e360c4c0` and reproduced: −11.90 / −3.65 / −3.64 / −3.90 % against
−11.47 / −3.35 / −3.73 / −3.74 % before, through a full reorganization of the tree.

Two things the new body reports that this one did not. The built-in control at T=4096 — the one width
where `full` applies and the change cannot act — had only two surviving passes in the wide sweep, so
it was re-taken on its own with eight: it reads −0.12 % inside its 0.77 % null while all eight
neighbours read −2.8 to −3.0 %. And the sparse cell is now reported as unresolved rather than as a
result: −0.64 % here against exactly 0.00 % before, both inside their own nulls.

Issue #247, opened 2026-09-15
NVFP4 W4A4: activation scales are staged as BlockM separate 16-byte TMA requests


MichaelDementii:
**Level of the claim: operator.** NVFP4 W4A4 Linear and the three fused projections that share the
route.

## The problem

`make_nvfp4_w4a4_tma_descriptors` describes the activation scale plane as `kGroupsPerRow` columns by
`tokens` rows and asks for a box of 16 columns by `BlockM` rows. The plane is row-major by token, so
those 16 bytes are `BlockM` separate rows at the plane's row stride: one pipeline stage delivers
4 KiB of scales as `BlockM` distinct 16-byte transactions, while the activation codes beside it in
the same stage travel as one 64-byte box per row.

The comment at that call site records the constraint honestly — "TMA's innermost box is at least one
16-byte transaction" — but the narrowness is a property of the *layout*, not of TMA: nothing requires
the plane to be row-major, because only the quantizer writes it and only this route reads it.

## Expected behaviour

Writing the same bytes tile-contiguous, one `[BlockM tokens, 16 groups]` tile per request, makes the
request as wide as the tile. The byte order inside a tile is unchanged, so the shared-memory image
the MMA reads is identical and the consumer side does not move.

## Affected contract and ownership boundary

The scale plane is a private workspace between `launch_nvfp4_w4a4_quantize` and the W4A4 GEMM; no
public type, tensor layout or workspace size changes. Its extent is unchanged: the tiled layout is a
bijection onto the same byte range, for a whole number of token tiles.

The layout must agree with the route, and after `refactor(ops): organize nvfp4 execution by matrix
shape` the route condition lives in each shape's `select_a4` and is no longer uniform —
`perf(ops): tune nvfp4 34816x5120 linear routes` put that shape at 256 while the other four stayed at
1024. So the layout has to follow whatever the shape actually selects, rather than a shared
predicate.

## Measurement

`ninfer_linear_bench --qtype NVFP4 --policy a4 --execution graph --warmup 20 --repeat 300
--flush-mib 256`, RTX 5090 `sm_120a`, CUDA 13.1, Release, base `e360c4c0`. Three arms from one tree:
master, master with the change, and the master binary again under a second label as an
identical-baseline null. Four passes, arm order rotated, paired inside a pass; a pass whose two
identical binaries disagree by more than 1 % in a cell is discarded for that cell (2 of 120 were).

Candidate against master, per registered shape and token count:

[...]

PR #248, opened 2026-09-15
NVFP4 W4A4: read activation scales one tile per TMA request instead of BlockM 16-byte requests

MichaelDementii:
## Problem and scope

Related Issue: #247

`make_nvfp4_w4a4_tma_descriptors` describes the activation scale plane as `kGroupsPerRow` columns by
`tokens` rows and asks for a box of 16 columns by `BlockM` rows. The plane is row-major by token, so
those 16 bytes are `BlockM` separate rows at the plane's row stride: one pipeline stage delivers
4 KiB of scales as `BlockM` distinct 16-byte transactions, while the activation codes beside it in
the same stage travel as one 64-byte box per row.

The narrowness is a property of the layout, not of TMA. Only the quantizer writes that plane and
only this route reads it, so it does not have to be row-major.

Scope is the W4A4 activation scale plane and the code that writes and reads it: the quantizer, both
W4A4 TMA descriptor builders, and the A4 route selection that has to keep the two in step. No public
type, tensor layout, workspace size or dispatch threshold changes.

## Implementation

The quantizer writes the plane in `[kNvfp4TmaBlockM tokens, kNvfp4ScaleTileGroups groups]` tiles when
the GEMM that follows takes the TMA route, and row-major otherwise; the descriptor addresses one tile
per request.

**The shared-memory image is unchanged, byte for byte.** The box is `BlockM` bytes wide and the
tiled plane's row stride is also `BlockM`, so a request covers a contiguous 4 KiB range and TMA
copies it without reordering. A scale for `(token, group)` therefore lands at the same offset inside
the stage as before, and the consumer side of the kernel is not touched. The plane's extent is also
unchanged — for K=6144 and T=1024, 96 tiles × 4096 B = 393 216 B = `tokens × K/16`, exactly what
`allocate_nvfp4_w4a4_workspace` reserves.

**The layout and the route travel together.** They must agree — a row-major plane read as tiled is
silently wrong numbers, not a failure — so `select_a4` now returns `Nvfp4A4Route`, the GEMM plus the
layout it reads, and the two kinds of route are constructed in one place each
(`nvfp4_a4_tma_route` / `nvfp4_a4_mma_route`). A shape selects a route rather than selecting the two
halves separately, so pairing them wrongly is not expressible.

**The per-shape floors stay per shape.** `34816x5120` routes from 256 and the other four from 1024;
the three fused projections keep their own predicate as well. Collapsing those into one shared
condition would be smaller, but this route has just demonstrated that the floors are tuned per call
site, and a shared predicate silently mis-selects the layout the first time one of them moves again.

Supporting changes, all of them consequences of the above:

* `make_nvfp4_w4a4_tma_descriptors` pins `BlockM == kNvfp4TmaBlockM`. It previously admitted 128,
  which no registered schedule uses and which would have addressed those tiles mis-shaped. The
  schedule itself keeps master's 128/256 parameterization; the requirement belongs to the descriptor.
* `Nvfp4W4a4TmaSchedule` now asserts the two facts the tile width depends on — that a tile is one
  shared-memory row per token, and that it spans two K tiles, which is why the producer fetches it on
  even k-tiles only.
* The whole-tile precondition is checked where the layout is acted on, and the tiled instantiation is
  gated on a K whose group count divides into whole tiles, so a future K that does not is a runtime
  message rather than a compile error across the quantizer.
* Removing the four duplicated local `kTmaBlockM` constants re-aligns the `using` blocks around them;
  about fifteen lines of the diff are that whitespace.

## Verification

**Numerical.** A layout disagreement would show as wrong values, not as a failure, so the check is
byte equality of the reported statistics against master on identical test widths. Both arms built
from one tree, `NINFER_OP_REPORT_STATS=1` at `%.17g`, across the five Op families that share the
plane:

```
ctest --test-dir build -j 4                                    # 118/118 on both arms
NINFER_OP_REPORT_STATS=1 ./ninfer_linear_nvfp4_a4_test          # 262 records
NINFER_OP_REPORT_STATS=1 ./ninfer_linear_add_nvfp4_test         #  28
NINFER_OP_REPORT_STATS=1 ./ninfer_linear_swiglu_nvfp4_test      #  29
NINFER_OP_REPORT_STATS=1 ./ninfer_attn_input_proj_test          # 2832
NINFER_OP_REPORT_STATS=1 ./ninfer_gdn_input_proj_test           # 134
```

**3285 numeric records, byte-identical.** The gate is shown able to fail: reversing the group-tile
order inside `nvfp4_tiled_scale_offset` makes `ninfer_linear_nvfp4_a4_test` fail at
`[5120,17408] T=1024`.

Route boundaries are new coverage, because the new failure mode lives exactly there: 1023/1024/1025
for `linear_add`, `attn_input_proj` and `gdn_input_proj`, and 255/256/257/512 for the fused SwiGLU
route, which has its own floor. Adding 1025 exposed a pre-existing defect in the `linear_add`
harness — it sized its host buffers from a literal 1024 while the invocation list drives them, so any
width past that read past the end. The bound is now taken from the list. That is fixed here because
the new width cannot be added without it.

**Operator.** `ninfer_linear_bench --qtype NVFP4 --policy a4 --execution graph --warmup 20
--repeat 300 --flush-mib 256`, four passes with the arm order rotated, paired inside a pass, third
arm the master binary again as an identical-baseline null; 1 of 120 passes discarded on that null.

| shape | 256 | 512 | 768 | 1024 | 1536 | 2048 |
|---|--:|--:|--:|--:|--:|--:|
| 5120×6144 | 0.00 | 0.00 | 0.00 | **−9.20** | **−4.71** | **−8.76** |
| 5120×17408 | −0.36 | 0.00 | −0.23 | **−14.17** | **−6.45** | **−10.35** |
| 14336×5120 | 0.00 | 0.00 | 0.00 | **−6.23** | **−9.34** | **−7.75** |
| 16384×5120 | 0.00 | −0.01 | 0.00 | **−2.08** | **−8.66** | **−3.04** |
| 34816×5120 | **−2.15** | **−2.58** | **−9.36** | **−8.66** | **−9.60** | **−9.03** |

Eighteen routed cells of eighteen are faster, median −8.7 %, worst −2.1 %. The twelve cells at or
near 0.00 % are widths below their own shape's floor: they do not take this route, so they are
controls rather than results, and the largest of them (−0.36 %) sits inside its own 0.41 % null.

**Per kernel**, since the write side changes too — a warp's 32 consecutive scale stores were one
contiguous 32-byte run and are now two 16-byte runs 4 KiB apart. `nsys --cuda-graph-trace=node`,
median over 45 launches:

| cell | kernel | master | this change |
|---|---|--:|--:|
| 5120×17408, T=1024 | `nvfp4_w4a4_tma_kernel` | 140.799 µs | **117.598 µs** (−16.5 %) |
| | `nvfp4_w4a4_quantize_kernel` | 28.415 µs | 28.543 µs (**+0.45 %**) |
| 34816×5120, T=256 | `nvfp4_w4a4_tma_kernel` | 88.671 µs | **87.008 µs** (−1.9 %) |
| | `nvfp4_w4a4_quantize_kernel` | 3.840 µs | 3.840 µs (0.00 %) |

The quantizer costs 128 ns more where the GEMM saves 23.2 µs, and nothing measurable at the narrow
width.

**Absolute performance of the resulting implementation**, logical bandwidth against the benchmark's
own references — 1792 GB/s nominal, and 1674.5 GB/s sustained read, labelled separately:

| shape | T | µs | logical GB/s | % 1792 | % 1674.5 | TFLOP/s |
|---|--:|--:|--:|--:|--:|--:|
| 34816×5120 | 256 | 93.408 | 1292.4 | 72.12 | 77.18 | 977.1 |
| 34816×5120 | 512 | 154.880 | 911.4 | 50.86 | 54.43 | 1178.6 |
| 34816×5120 | 1024 | 258.976 | 703.0 | 39.23 | 41.98 | 1409.7 |
| 14336×5120 | 1024 | 124.080 | 653.9 | 36.48 | 39.05 | 1211.5 |
| 16384×5120 | 1024 | 152.752 | 597.2 | 33.33 | 35.66 | 1124.7 |
| 5120×6144 | 1024 | 60.640 | 672.2 | 37.51 | 40.14 | 1062.4 |
| 5120×17408 | 1024 | 148.704 | 647.4 | 36.13 | 38.66 | 1227.5 |

## What is not claimed, and what was not run

* **No Tensor Core utilisation for this path.** `linear_bench` defines dense peaks for fp8 and bf16
  only and prints a dash in `TC_profile`/`TC_%` for NVFP4 rows; the measured logical throughput here
  exceeds every constant it defines, so the denominator is not established anywhere I can cite.
  Establishing one is worth doing separately.
* The hot interval `1 <= T <= 128` and the `T=512` anchor of the four shapes that route from 1024 are
  untouched by this change — those widths do not select this route — and the table above shows them
  reading zero rather than assuming it. `T=512` **is** affected for `34816x5120`, which routes from
  256, and is reported separately above.
* No end-to-end measurement. The claim is at the operator level.
* `linear_bench` was swept at the six widths above per shape, not at every multiple of 256.


MichaelDementii:
Supersedes #238, which was written against the NVFP4 layout before the per-shape reorganization. Scope is in #247.

@codex review

Neroued (approved):

Issue #249, opened 2026-09-15
NVFP4 fused SwiGLU: the TMA epilogue pays for an accurate silu before rounding to bf16


MichaelDementii:
**Level of the claim: operator**, with the same kernel's share of a real prefill reported alongside.

## The problem

The fused NVFP4 SwiGLU TMA epilogue computes `silu(gate * alpha) * (up * alpha)` and rounds the
product to bf16 on the next instruction. `silu` is `x / (1.0f + expf(-x))`: `expf` compiles to a
guarded slow path and the divide to a Newton refinement, and both are paid in full for a value that
is about to lose 16 mantissa bits.

## Expected behaviour

An activation written for a caller that rounds immediately:

```cpp
__device__ __forceinline__ float silu_approx(float x) {
    const float e = __expf(-fabsf(x));
    const float r = __fdividef(1.0f, 1.0f + e);
    return (x >= 0.0f ? x : x * e) * r;
}
```

The exponential is folded onto the side that cannot overflow, so the divisor stays in `[1, 2]` where
`__fdividef` carries its documented 2 ulp.

Measured on this base, `ninfer_nvfp4_linear_swiglu_bench --policy a4 --warmup 20 --repeat 300`, four
passes with the arm order rotated and the master binary again as an identical-baseline null:

| T | route | change |
|--:|---|--:|
| 129 | other | +0.01 % |
| 255 | other | +0.00 % |
| **256** | fused TMA | **−10.55 %** |
| **512** | fused TMA | **−13.38 %** |
| **768** | fused TMA | **−11.61 %** |
| **1024** | fused TMA | **−10.84 %** |
| **1536** | fused TMA | **−10.73 %** |
| **2048** | fused TMA | **−9.97 %** |

Six routed widths of six are faster, median −10.8 %; the two widths that take other routes read zero,
which is the control. No passes were discarded.

In a real prefill the same kernel is **17.9 % of all kernel time** (Qwen3.8-27B NVFP4, 8192-token
prompt, prefill chunk 1024, 1344 launches per pass) and gains **2.9 %** there, median of three passes,
range 1.1 to 3.3 %. The operator benchmark flushes 256 MiB of L2 before every sample and the model
runs warm and back-to-back; I did not establish which difference in conditions accounts for the gap
between −10.8 % and −2.9 %, so both are stated with their conditions rather than reconciled.

## The accuracy question, which is the real one

**This is not bitwise identical**, so the trade has to be measured rather than asserted. Against a
[...]

PR #250, opened 2026-09-15
NVFP4 fused SwiGLU: approximate the activation in the TMA epilogue, which rounds to bf16 immediately

MichaelDementii:
## Problem and scope

Related Issue: #249

The fused NVFP4 SwiGLU TMA epilogue computes `silu(gate * alpha) * (up * alpha)` and rounds the
product to bf16 on the very next instruction. `silu` is `x / (1.0f + expf(-x))`, where `expf`
compiles to a guarded slow path and the divide to a Newton refinement — both paid in full for a value
about to lose 16 mantissa bits.

Scope is one call site: the epilogue of `nvfp4_linear_swiglu_w4a4_tma`. At least nine other epilogues
in the tree compute `silu` and round straight to bf16; none is touched here, because each would need
its own accuracy argument.

## Implementation

`silu_approx` in `ops/common/math.cuh`, beside `silu`:

```cpp
__device__ __forceinline__ float silu_approx(float x) {
    const float e = __expf(-fabsf(x));
    const float r = __fdividef(1.0f, 1.0f + e);
    return (x >= 0.0f ? x : x * e) * r;
}
```

The exponential is folded onto the side that cannot overflow. `1 + exp(-x)` exceeds 2^126 for x below
−87.34 and `__fdividef` returns zero past that, so the unfolded form loses the tail; written this way
the divisor stays in `[1, 2]`, where `__fdividef` carries its documented 2 ulp. `__expf` rather than
the tree's `exp2_approx`, because the documented error bound belongs to the intrinsic and folding the
`log2(e)` scale in by hand would put a rounding of its own in front of it.

Two files, +22/−4.

## Verification

### Accuracy, which is the load-bearing part

**This is not bitwise identical**, so the trade is measured rather than asserted. A dense sweep
against a double-precision reference counts how often each form rounds to a different bf16 than the
mathematics:

| range, points | `silu` (master) | `silu_approx` |
|---|--:|--:|
| [−20, 20], 40 000 001 | 970 — **1 in 41 237**, max rel 1.10e−06 | 1285 — **1 in 31 128**, max rel 1.86e−06 |
| [−8, 8], 40 000 001 | 465 — 1 in 86 022, max rel 3.76e−07 | 646 — 1 in 61 920, max rel 7.55e−07 |

In the band the model actually uses the approximate form disagrees with the mathematics about
1.3× more often. **That is the cost, and it is the first thing to weigh.**

In the tail the ordering reverses, by much more:

| range | `silu` (master) | `silu_approx` |
|---|--:|--:|
| [−100, 100], 40 001 points | **1761** differ, max rel **1.000** | **16** differ, max rel 1.80e−02 |
| [−95, −75], 4001 points | 1257 differ | **1** |
| returns exactly zero where the reference is a normal bf16 | **628**, first at x = −91.86 | **0** |

Below about −87.3 the accurate form's denominator overflows in float and it returns zero while the
reference is still representable; folding the exponential onto the non-overflowing side removes that.

`ctest` 118/118, unchanged from master.

### Operator

`ninfer_nvfp4_linear_swiglu_bench --policy a4 --warmup 20 --repeat 300`, four passes with the arm
order rotated, paired inside a pass, third arm the master binary again as an identical-baseline null.
The fused TMA route runs at multiples of 256 from 256 up; 129 and 255 take other routes, so they are
controls rather than results.

| T | route | change | null p95 |
|--:|---|--:|--:|
| 129 | other | +0.01 % | 0.022 |
| 255 | other | +0.00 % | 0.000 |
| **256** | fused TMA | **−10.55 %** | 0.272 |
| **512** | fused TMA | **−13.38 %** | 0.190 |
| **768** | fused TMA | **−11.61 %** | 0.000 |
| **1024** | fused TMA | **−10.84 %** | 0.111 |
| **1536** | fused TMA | **−10.73 %** | 0.026 |
| **2048** | fused TMA | **−9.97 %** | 0.238 |

Six routed widths of six are faster, median −10.8 %, worst −9.97 %. No passes were discarded.

Absolute performance of the resulting implementation, from the same runs:

| T | µs | logical GB/s | TFLOP/s |
|--:|--:|--:|--:|
| 256 | 115.936 | 964.4 | 787.2 |
| 512 | 175.344 | 703.4 | 1041.0 |
| 1024 | 308.448 | 474.7 | 1183.6 |
| 2048 | 579.072 | 332.5 | 1260.9 |

### The same kernel in a real prefill

Qwen3.8-27B NVFP4, 8192-token prompt, prefill chunk 1024, `nsys --cuda-graph-trace=node`, three
passes per arm, 1344 launches of this kernel per pass. The artifacts had to be upgraded with
`tools/upgrade_ninfer_v2_to_v3.py` first.

| | master | this change |
|---|--:|--:|
| share of all kernel time | **17.9–18.0 %** | 17.4 % |
| total time in this kernel | 452.9 / 453.8 / 454.5 ms | 439.8 / 438.7 / 449.4 ms |

Median **−2.9 %**, range −1.1 to −3.3 %, so roughly **0.5 % of prefill kernel time**.

**The operator figure and the model figure disagree and are not reconciled here.** The benchmark
flushes 256 MiB of L2 before every sample; the model runs warm and back-to-back. I did not establish
which difference in conditions accounts for −10.8 % against −2.9 %, so both are given with their
conditions rather than averaged or explained away.

## What is not claimed

* No end-to-end throughput number. The claim is the operator and this kernel's share of prefill.
* No Tensor Core utilisation: `linear_swiglu_bench` reports logical bandwidth and TFLOP/s but the
  tree defines no dense NVFP4 tensor peak to divide by.
* The nine other `silu` call sites are untouched and unmeasured.
* The accuracy sweep is over a dense grid of the input range, not over activations captured from a
  running model.


MichaelDementii:
Supersedes #239. Scope is in #249.

@codex review

Neroued:
Local review found no blocking implementation issues. I recommend merging after the attribution cleanup below.

Tested `89547367` on top of master `1d8587bc`, using an RTX 5090 and CUDA 13.1:

- The complete NVFP4 LinearSwiGLU FP64-oracle test passed.
- Supplemental checks of `SiLU(gate) × up → BF16` confirmed small rounding differences within the observed numerical limits.
- Three alternating benchmark passes, with 20 warmups, 300 samples and a 256 MiB L2 flush, showed **10.2–15.1% lower Op latency** across the six affected widths, with a **14.4% median reduction**.
- Formatting and diff checks passed.

**Before merging, please remove `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>` from the commit message and ensure it is absent from the final squash message. Preserve the human author attribution.**

If this lands before #253, please recalibrate #253’s SwiGLU routing against the updated master, since this change alters the fused implementation’s cost.

MichaelDementii:
Attribution cleaned: the commit is now `65fab20e`, same author, no `Co-Authored-By` trailer. I have also dropped the "Generated with" line from our PR bodies going forward — you did not ask for that one, but it is the same signal in a different place.

Noted on the ordering: if this lands first, #253's SwiGLU routing gets recalibrated against the master that contains it, since this change alters the fused implementation's cost. That recalibration is what I am working on now, and I would rather do it once, after this is in.

Neroued (approved):

Issue #252, opened 2026-09-15
NVFP4 W4A4: the TMA route requires a token count that is a multiple of 256


MichaelDementii:
**Level of the claim: operator.** NVFP4 W4A4 Linear and the three fused projections, plus the fused
SwiGLU route.

## The problem

Every W4A4 TMA route condition requires `tokens % 256 == 0`:

```
src/ops/linear/nvfp4/shapes/n14336_k5120.cu   tokens >= 1024 && tokens % 256 == 0
src/ops/linear/nvfp4/shapes/n16384_k5120.cu   tokens >= 1024 && tokens % 256 == 0
src/ops/linear/nvfp4/shapes/n34816_k5120.cu   tokens >=  256 && tokens % 256 == 0
src/ops/linear/nvfp4/shapes/n5120_k6144.cu    tokens >= 1024 && tokens % 256 == 0
src/ops/linear/nvfp4/shapes/n5120_k17408.cu   tokens >= 1024 && tokens % 256 == 0
  … and the same condition in linear_add, gdn_input_proj and attn_input_proj
```

The requirement is the M tile width, not anything about the arithmetic. A prefill chunk only has to
be a multiple of 128, and the final chunk of a prompt is whatever is left, so on most real widths the
route is simply not taken. `34816x5120` feels this hardest: it routes from 256, so a ragged width
falls all the way back to the MMA route.

## Expected behaviour

Let the last M tile be partial: ceiling-divide the grid, bound the kernel by the real token count,
pad the activation-scale plane up to whole tiles and fill the padding with zeroes. A padded row then
owns no input and no code byte; its codes arrive as TMA zero-fill and a zero scale times any E2M1
code is exactly zero, so it contributes nothing.

Measured on this base — arm A is the tiled-scale change alone, arm B that plus this one, third arm
the A binary again as an identical-baseline null, four passes with the order rotated,
`ninfer_linear_bench --qtype NVFP4 --policy a4 --execution graph --warmup 20 --repeat 300
--flush-mib 256`:

| shape | 1024 | **1025** | **1200** | 1280 | **1500** | 2048 |
|---|--:|--:|--:|--:|--:|--:|
| 5120×6144 | 0.00 | **−7.90** | **−11.19** | 0.00 | **−10.13** | +0.01 |
| 5120×17408 | 0.00 | **−10.20** | **−12.51** | 0.00 | **−12.70** | 0.00 |
[...]

PR #253, opened 2026-09-15
NVFP4 W4A4: take a partial last M tile, so the route stops needing a multiple of 256

MichaelDementii:
## Problem and scope

Related Issue: #252

Every W4A4 TMA route condition requires `tokens % 256 == 0` — the five registered Linear shapes and
`linear_add`, `gdn_input_proj`, `attn_input_proj`. That requirement is the M tile width, not anything
about the arithmetic. A prefill chunk only has to be a multiple of 128 and the final chunk of a prompt
is whatever is left, so on most real widths the route is not taken. `34816x5120` feels it hardest: it
routes from 256, so a ragged width falls all the way back to the MMA route.

Scope is the tail handling of that route. **The fused SwiGLU route is not moved** — see below.

## Implementation

partial. Three things follow:

* **The scale plane is padded to whole tiles.** The tiled layout addresses whole tiles, so the plane
  is allocated over `nvfp4_w4a4_padded_tokens` — at most 255 tokens of scales, under 0.3 MiB on the
  widest registered K — and the quantizer writes zeroes into the padding, which the consumer's box
  covers and would otherwise read from whatever the arena last held. A padded row owns no input and
  no code byte: its codes arrive as TMA zero-fill, and a zero scale times any E2M1 code is exactly
  zero.
* **That padding is the layout's new precondition, so it is checked.** The plane's extent now travels
  with the pointers and `launch_nvfp4_w4a4_quantize` verifies it. The whole-tile check it replaces
  was the only guard on the tiled contract; without a replacement, a caller that sized its own
  `Nvfp4W4a4Workspace` would silently write up to 255 tokens of scales past the end.
* **The epilogue bounds itself.** The activation *code* descriptor carries the real token count as its
  row extent, so TMA zero-fills the code rows past the end; the scale plane is padded instead. Either
  way a padded lane contributes nothing and only has to stay off other people's memory, so its token
  index is clamped and its store dropped.

### The fused SwiGLU route is deliberately left at whole tiles

An earlier revision of this PR also sent ragged widths of that route to the fused kernel. Review found
the public `linear` + `silu_mul` composition faster there, which reproduces: 11.35 / 9.47 / 7.71 % at
T=384 / 513 / 769 on my machine against the 11.6 / 9.8 / 8.5 % reported.

Measuring the two candidate conditions through the public Op shows the answer depends on whether #250
is in the tree:

| T | 384 | 513 | 769 | 1025 | 1500 | 2000 |
|---|--:|--:|--:|--:|--:|--:|
| without #250 | −11.35 | −9.47 | −7.71 | −6.55 | 0.00 | **+1.38** |
| with #250 | **+3.63** | **+5.88** | **+9.12** | **+9.26** | **+17.45** | **+17.89** |

Negative means the composition is faster. So there is no single condition that is right under both
merge orders, and shipping the ragged routing here would regress the ragged band if this landed
first. Leaving the route at whole tiles is correct either way — and those widths still gain, because
the composition's own `linear` now takes the TMA path. The routing change follows separately once
#250 lands, with its own measurement.

## Verification

RTX 5090 `sm_120a`, CUDA 13.1, Release, base `beedffa0`, one card, one session. Three arms, the third
being the master binary again under a second label as an identical-baseline null, four passes with the
arm order rotated, paired inside a pass.

### Linear

`ninfer_linear_bench --qtype NVFP4 --policy a4 --execution graph --warmup 20 --repeat 300
--flush-mib 256`

| shape | 1024 | **1025** | **1200** | 1280 | **1500** | 2048 |
|---|--:|--:|--:|--:|--:|--:|
| 5120×6144 | 0.00 | **−7.90** | **−11.19** | 0.00 | **−10.13** | +0.01 |
| 5120×17408 | 0.00 | **−10.20** | **−12.51** | 0.00 | **−12.70** | 0.00 |
| 14336×5120 | 0.00 | **−6.30** | **−17.46** | 0.00 | **−24.66** | −0.01 |
| 16384×5120 | 0.00 | **−18.36** | **−24.59** | +0.06 | **−24.04** | 0.00 |
| 34816×5120 | 0.00 | **−45.31** | **−48.14** | 0.00 | **−47.56** | 0.00 |

### The three fused projections

They change route at the same widths and carry their own benchmarks, so they were swept separately at
the same 20/300:

| T | `attn_input_proj` | `gdn_input_proj` | `linear_add` 6144 | `linear_add` 17408 |
|---|--:|--:|--:|--:|
| 1024 | +0.06 | +0.02 | 0.00 | 0.00 |
| **1025** | **−5.32** | **−17.28** | **−13.20** | **−19.42** |
| **1200** | **−16.37** | **−25.09** | **−5.03** | **−11.87** |
| 1280 | +0.01 | 0.00 | 0.00 | 0.00 |
| **1500** | **−23.72** | **−23.29** | **−13.31** | **−19.94** |
| 2048 | 0.00 | +0.23 | 0.00 | 0.00 |

Fifteen ragged cells of fifteen on Linear and twelve of twelve on the projections are faster, medians
**−17.5 %** and **−16.8 %**. The twenty-seven whole-tile cells take the same route on both arms and
read 0.00 %, so they are the control rather than a result. One pass of 120 and four of 96 were
discarded on the null.

### Output

`NINFER_OP_REPORT_STATS=1` at `%.17g` on both arms over identical test widths across the five Op
families that share the route: **3290 numeric records, all byte-identical**. Nothing in this revision
moves a route between two different implementations, so there is nothing that should differ.

`ctest` 120/120. `clang-format` clean.

Tests gain **1500** — five whole tiles and a mostly full sixth. The widths already present become
ragged rather than falling to MMA: 1025 on every shape and on the three projections, and 257/513/769
on the shape that routes from 256.

## Limits

* **The padding is not free.** At T=1025 four fifths of the last tile is padding, and that width gains
  least of the three ragged ones measured on three of the five shapes. It is still a gain everywhere
  measured.
* The scale plane grows by up to 255 tokens of scales, paid whichever layout the quantizer writes.
* Ragged widths were swept at three points per shape, not at every non-multiple of 256.
* No end-to-end number; the claim is at the operator level.
* No Tensor Core utilisation: the benchmark defines no dense NVFP4 tensor peak to divide by.


MichaelDementii:
Supersedes #240. Scope is in #252. Depends on #248, whose commit is the first of the two here.

@codex review

Neroued:
The tail support passes the local correctness checks and delivers substantial Op-level gains. Please address the SwiGLU route selection below before merging, and rebase/clean up the commit attribution.

This comment is posted by an agent on behalf of @Neroued. The maintainer has reviewed and authorized these requests.

Reviewed increment: `a79e2eb30bc51447b6c5ed8410d5eee5a428ed08`, applied without conflicts to master `1d8587bcfe850fba310d8833552f3c0c07e3a4bd`, which already contains #248. The duplicate #248 commit was not applied again.

**1. Recalibrate the newly admitted ragged SwiGLU routes.**

In `nvfp4_linear_swiglu_plan.cpp`, `kRaggedFloor = 384` and the `tokens >= kRaggedFloor` condition select the fused TMA implementation where the existing public `linear` + `silu_mul` composition is faster.

The initial production benchmark showed a jump from about 173 μs at T=383 to 196 μs at T=384. A follow-up comparison at identical widths confirmed the issue:

| T | Selected `linear_swiglu`, μs | Public `linear` + `silu_mul`, μs | Composition latency reduction |
|---:|---:|---:|---:|
| 383, control | 172.992 | 172.992 | 0.0% |
| 384 | 195.808 | 173.024 | 11.6% |
| 385 | 195.840 | 173.024 | 11.7% |
| 513 | 232.704 | 209.952 | 9.8% |
| 769 | 312.320 | 285.664 | 8.5% |

These are measurements from the same candidate build, using the same inputs and weights. The comparison harness was a temporary derivative of `ninfer_nvfp4_linear_swiglu_bench`: one mode calls public `linear_swiglu`; the other calls public `linear` into a BF16 `[34816,T]` projection and then public `silu_mul` on its gate/up row views. Both modes retain the same fixture allocations, with projection storage and workspace allocated outside timing. No private launcher or production implementation was modified.

Environment/method: RTX 5090, driver 610.74, CUDA 13.1.115, GCC 13.3, Release/sm_120a; eager CUDA-event timing of the complete Op/composition; 256 MiB L2 flush before each sample; 20 warmups and 300 samples. Three passes rotated production/composition/repeated-production order. Times above are medians of the per-pass medians; percentage changes use paired ratios. The repeated-production control differed by at most 0.115% at the four affected widths shown above. The token list was `255,257,383,384,385,511,512,513,767,768,769`. No profiling or model inference was used.

The PR is still faster than master at these points. The requested change is to avoid selecting a slower implementation within the newly expanded ragged domain when a faster existing composition is available. Please compare the complete alternatives, adjust the new dispatch condition accordingly, and include relevant boundary correctness and benchmark results. This does not require treating pre-existing whole-tile choices as a separate optimization project, or guessing a replacement threshold from this small table.

For context, the original five Linear shapes at T=1025/1200/1500 reproduced a median latency reduction of **17.26%**, versus 17.5% in the PR. All five affected test executables passed on both baseline and candidate, including the new T=1500 and SwiGLU 383/384/385 cases. Formatting and diff checks passed. The core tail-handling change is worth keeping.

**2. Rebase onto current master and drop the already merged #248 dependency.**

#248 was squash-merged. The PR branch still includes its original commit and currently reports merge conflicts. The tail-only increment applies cleanly to current master; please update the branch so it contains only the remaining change.

**3. Remove the AI co-author trailers.**

Both commits currently include `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>`. Please remove that trailer from the remaining PR commit(s), preserve the human author attribution, and ensure it is absent from the final merge/squash message. The maintainer does not want AI tools listed as commit authors or co-authors.

MichaelDementii:
You were right, and the recalibration went further than adjusting the threshold: the SwiGLU routing
is now out of this PR entirely.

## Your cell, reproduced

Same comparison through the public Op, four passes, third arm the same binary again as a null:

| T | you | me |
|---:|---:|---:|
| 384 | 11.6 % | **11.35 %** |
| 513 | 9.8 % | **9.47 %** |
| 769 | 8.5 % | **7.71 %** |

Different harness, same conclusion. Your T=383 control is the part that told me where I had gone
wrong: both arms read 172.992 there because both take the composition — and that composition already
contains this PR's faster `linear`. The change made the alternative faster and then routed away from
it. I had compared the fused route against master and against the baseline *as it was*, never against
what the dispatcher would choose *after* the change landed.

## Why the threshold is gone rather than moved

Measuring the two candidate conditions across 257..2100 shows the answer depends on whether #250 is
in the tree:

| T | 384 | 513 | 769 | 1025 | 1500 | 2000 |
|---|--:|--:|--:|--:|--:|--:|
| without #250 | −11.35 | −9.47 | −7.71 | −6.55 | 0.00 | **+1.38** |
| with #250 | **+3.63** | **+5.88** | **+9.12** | **+9.26** | **+17.45** | **+17.89** |

Negative means the composition is faster. Two things follow.

The `kRaggedFloor = 384` I shipped was an invention. With #250 in the tree the fused route is at
least as fast from 257 upward — 257 is a wash, 280 is +1.5 %, 448 is +6.7 % — so there is no floor to
find. I had excluded 257..383 by arguing that a second, nearly empty tile is a third of the work,
without noticing that the composition's own GEMM runs the same ragged grid with the same partial
tile. Padding is not what separates them.

And there is no condition that is right under both merge orders. I had intended to give you both
tables and let the order decide, but that would have been offering a choice between one good outcome
and one worse than the revision you already rejected: without #250 the fused kernel needs two tiles
from 257 onward, so the band 257..383 would regress by roughly 13–18 %.

So this PR now carries only the mechanism, and `resolve_route` is back to master's whole-tile
condition. That is correct under either order, and the ragged widths still gain, because the
composition's `linear` takes the TMA path. The routing change follows as its own one-line PR once
#250 lands, measured against the master that contains it.

## The six other sites that change route

Your point applies beyond SwiGLU: `linear_add`, `gdn_input_proj` and `attn_input_proj` also start
taking ragged widths, and my measurement had covered Linear only. Swept at the same 20/300:

| T | `attn_input_proj` | `gdn_input_proj` | `linear_add` 6144 | `linear_add` 17408 |
|---|--:|--:|--:|--:|
| 1024 | +0.06 | +0.02 | 0.00 | 0.00 |
| **1025** | **−5.32** | **−17.28** | **−13.20** | **−19.42** |
| **1200** | **−16.37** | **−25.09** | **−5.03** | **−11.87** |
| 1280 | +0.01 | 0.00 | 0.00 | 0.00 |
| **1500** | **−23.72** | **−23.29** | **−13.31** | **−19.94** |
| 2048 | 0.00 | +0.23 | 0.00 | 0.00 |

Twelve ragged cells of twelve faster, median −16.8 %; the whole-tile rows are controls and read zero.

## One thing I fixed that you did not ask about

The earlier revision deleted the whole-tile check in `launch_nvfp4_w4a4_quantize` and replaced it with
nothing. With padding, the quantizer writes past the real token count and simply assumed the caller
had allocated through `allocate_nvfp4_w4a4_workspace`; `Nvfp4W4a4Workspace` is two raw pointers, so a
caller sizing its own would have had up to 255 tokens of scales written past the end — 81.6 KB at
K=5120, silently. The plane's extent now travels with the pointers and is checked where the old check
lived.

## The rest

Rebased onto `beedffa0`; the branch is one commit and no longer carries #248. The `Co-Authored-By`
trailer is gone from it and from the other open PRs, and I have dropped the "Generated with" line
from our PR bodies as well — you did not ask for that one, but it is the same signal in a different
place.

Since nothing in this revision moves a route between two different implementations, the output is now
unchanged everywhere: 3290 numeric records byte-identical across the five Op families, `ctest`
120/120.

Neroued (approved):

Issue #254, opened 2026-09-15
SparseMoe small-T: the Q5 routed-down Rows2 window stops at 5, well short of its measured crossover


MichaelDementii:
**Level of the claim: operator.** SparseMoe small-T D4, decode path.

## The problem

`resolve_sparse_moe_small_t_plan` picks the D4 row split from a per-codec token threshold. For
`Q5_G64_FP16` that threshold is 5:

```cpp
plan.d4_schedule = tokens <= 2   ? Rows1
                   : tokens <= 5 ? Rows2
                                 : Rows4;
```

`Rows` is only a grid split — `(kHidden / Rows, tokens)` — and the measured crossover on that codec
is much further right, so widths 6..17 run the slower schedule.

The threshold is per codec because the routed-down format is stored per layer in the artifact, so
both branches of this plan run inside the same decode round. On the 35B artifact in hand:

```
cd tools && python3 -m artifact.inspect qwen3_6_35b_a3b.ninfer
formats: {... 'q5_g64_fp16': 91, 'q6_g64_fp16': 5 ...}
```

## Measurement

`ninfer_sparse_moe_bench --distribution same --execution graph --cache cold --warmup 20 --repeat 300
--flush-mib 256`, RTX 5090 `sm_120a`, CUDA 13.1, Release, base `e360c4c0`. Five passes with the arm
order rotated; the third arm is the master binary again as an identical-baseline null. The
concentrated distribution — every token sharing one set of eight experts — is where `Rows2` is
weakest, so it is the binding case.

Every integer, not a sample of them:

| T | 3–5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18–20 |
|---|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| Q5 | 0.00 | **−3.50** | **−3.17** | **−2.90** | **−2.60** | **−2.38** | **−2.26** | −0.03 | **−1.92** | 0.00 | 0.00 | **−1.61** | **−1.54** | 0.00 |
| Q6 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 |

**No width in 6..17 regresses**, and 16 and 17 are among the gaining ones. One pass of 180 was
discarded on the null.

[...]

PR #255, opened 2026-09-15
SparseMoe small-T: widen the Q5 routed-down Rows2 window 5 to 17, its measured crossover

MichaelDementii:
## Problem and scope

Related Issue: #254

`resolve_sparse_moe_small_t_plan` picks the D4 row split from a per-codec token threshold. For
`Q5_G64_FP16` that threshold is 5, while the measured `Rows2`/`Rows4` crossover on that codec is much
further right, so widths 6..17 run the slower schedule.

Scope is that one threshold. `Q6_G64_FP16` keeps the 11 it has, and nothing else in the plan moves.

The threshold is per codec because the routed-down format is stored per layer in the artifact, so
both branches run inside the same decode round — on the 35B artifact, 91 tensors are `q5_g64_fp16`
and 5 are `q6_g64_fp16` (`cd tools && python3 -m artifact.inspect <artifact>`).

## Implementation

One number: the `Q5_G64_FP16` window goes **5 → 17**. `Rows` is only a grid split,
`(kHidden / Rows, tokens)`, and both instantiations already exist for both codecs, so nothing but
block geometry moves. One file, +13/−3, of which eight lines are the comment recording the
measurement.

## Verification

### Operator

```
ninfer_sparse_moe_bench --codec q4-q5 --distribution same --tokens T \
    --execution graph --cache cold --warmup 20 --repeat 300 --flush-mib 256
```

Five passes with the arm order rotated, paired inside a pass; the third arm is the master binary
again under a second label as an identical-baseline null. The concentrated distribution — every token
sharing one set of eight experts — is where `Rows2` is weakest, so it is the binding case. Every
integer was measured, not a sample:

| T | 3–5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18–20 |
|---|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| Q5 | 0.00 | **−3.50** | **−3.17** | **−2.90** | **−2.60** | **−2.38** | **−2.26** | −0.03 | **−1.92** | 0.00 | 0.00 | **−1.61** | **−1.54** | 0.00 |
| Q6 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 |

**No width in 6..17 regresses**, which is the claim the threshold rests on, and 16 and 17 are among
the gaining widths rather than the marginal ones. One pass of 180 was discarded on the null.

Three controls. T=3..5 select `Rows2` on both arms and T=18..20 select `Rows4` on both, so both ends
of the moved interval read zero. The Q6 branch is not touched at all: **eighteen cells of eighteen
read 0.00 %**, which is what shows the instrument separates changed from unchanged.

**Why three cells inside the range read zero.** Every absolute median is an exact multiple of
2.048 µs — 38.912, 45.056, 51.200, 57.344, 63.488, 69.632, 77.824, 90.112, 98.304, 104.448, 112.640,
118.784, 124.928, 131.072, 139.264, 145.408, 151.552, i.e. 19, 22, 25, 28, 31, 34, 38, 44, 48, 51,
55, 58, 61, 64, 68, 71 and 74 steps. The timer quantizes, and at T=12 one step is 2.08 % of the
operation. A cell reading 0.00 % means both arms landed on the same step, not that the two schedules
are equal there — which is why 12, 14 and 15 read zero between neighbours that gain.

Absolute performance of the resulting implementation, against the 1792.1 GB/s reference the benchmark
prints in its own header:

| T | 6 | 8 | 12 | 16 | 17 |
|---|--:|--:|--:|--:|--:|
| µs | 57.344 | 69.632 | 98.304 | 124.928 | 131.072 |
| unique-weight GB/s | 328.1 | 270.2 | 191.4 | 150.6 | 143.5 |
| share of that reference | 18.3 % | 15.1 % | 10.7 % | 8.4 % | 8.0 % |
| logical TFLOP/s | 6.03 | 6.63 | 7.04 | 7.39 | 7.48 |

### Output

Bit-exact by construction: `Rows` changes only how output rows are distributed over blocks, each row
accumulates over the intermediate dimension in the same order, and the nine paths are summed in rank
order in the epilogue, independent of `Rows`.

Witnessed where the change acts. `--spec dflash --draft-tokens 11` and `15` put decode at T=12 and
T=16, inside the moved window, so master runs `Rows4` and this branch runs `Rows2` on every decode
step. Three prompts × two widths, greedy, 35B artifact: **6 of 6 completions byte-identical**,
448–604 bytes each — the size is reported because two empty captures also compare equal.

That gate is shown able to fail. A third arm with the `Rows2` epilogue scaled by 1.001 — the schedule
this change selects, and nothing else — **differs from master in 4 of the same 6 cells**. Two cells
absorb a perturbation that small, so the witness is sensitive to this rather than to everything; it
also establishes that the branch really executes `Rows2` at those widths, since a `Rows2`-only
perturbation could not otherwise change the output.

`ctest` 118/118. `clang-format` clean.

## Limits

* One machine. An earlier campaign on a faster card put the Q5 crossover further right, at 19; this
  proposal takes the lower reading. That card is no longer available, so the figure is context, not
  evidence, and the threshold here rests on the table above alone.
* Readings below one 2.048 µs step cannot be resolved, so the zeros inside the range mean "not
  distinguishable", not "proven equal".
* Concentrated routing is the binding distribution and the only one swept densely here.
* No end-to-end number. At batch 1 the decode round runs T=4, inside the old window as well as the
  new one; T=6..17 needs batch ≥ 2 or speculation, and this card cannot resolve a per-layer operator
  effect of this size end to end.
* An earlier revision of this proposal also raised Q6 and was rejected for a 1.61 % regression at
  Q6/concentrated/T=17. Q6 is untouched here and reads zero at every width measured.


MichaelDementii:
Supersedes #242. Scope is in #254.

@codex review

Neroued (approved):

Issue #256, opened 2026-09-15
Q8 RowSplit MMA: the predicated staging path inherits its cache policy instead of choosing it


MichaelDementii:
**Level of the claim: operator.** Q8 RowSplit MMA GEMM and the Ops that share its tile.

## The problem

`q8_rowsplit_gemm_mma_kernel` stages activations, codes and scales one way on full tiles and another
on predicated ones:

```cpp
if constexpr (Full) {  cp_async<16, Cache::cg>(...);   // chosen
} else {               cp_async_zfill<16>(...);        // Cache::ca, the memory.cuh default
```

The predicated side never chose `ca`; it inherited `cp_async_zfill`'s default. Which side runs is
decided per launch, not per tile — `full` needs `(w.n % BM) == 0 && (x.ne[1] % BN) == 0 && ...` — so
any token count that is not a multiple of `BN` sends the whole GEMM down the predicated path with a
policy nobody picked.

## Expected behaviour

Make the policy a schedule parameter defaulting to `Cache::ca`, so adding it changes no existing
instantiation, and let one schedule opt in where it has been measured.

`ninfer_linear_bench --qtype Q8 --execution graph --warmup 20 --repeat 300 --flush-mib 256`, RTX 5090
`sm_120a`, CUDA 13.1, Release, base `e360c4c0`. Three passes with the arm order rotated, third arm the
master binary again as an identical-baseline null:

| shape and range | cells | median | worst |
|---|--:|--:|--:|
| feature 5120×25600, 129..255 | 10 | **−11.90 %** | −11.54 |
| generic 9216×2048, 4096..4223 | 122 | **−3.65 %** | −0.10 |
| mid 34816×5120, 129..319 | 11 | **−3.64 %** | −2.61 |
| lm head 248320×5120, 129..319 | 11 | **−3.90 %** | −3.12 |
| sparse 2048×4096, 896..1096 | 11 | −0.64 % | 0.00 |
| feature 50..56 (a different kernel) | 7 | **0.00 %** | 0.00 |
| feature 57..64 (BM=16) | 8 | **0.00 %** | +0.37 |
| feature 65..127 (BM=32) | 7 | **0.00 %** | 0.00 |

The last three rows are routes the proposal deliberately leaves alone. The sparse row is **not
resolved**: it read −0.64 % here and exactly 0.00 % on an earlier base, both inside their own nulls.

**The control this sweep contains.** `full` needs `(T % BN) == 0`, so T=4096 is the one width the
[...]

PR #257, opened 2026-09-15
Q8 GEMM: choose the predicated cache policy instead of inheriting it, operator -3.6 to -3.9% on ragged token counts

MichaelDementii:
## Problem and scope

Related Issue: #256

In `q8_rowsplit_gemm_mma_kernel` the full-tile branch spells `Cache::cg` and the predicated branch a
few lines below it does not spell anything — it inherits `Cache::ca` from `cp_async_zfill`'s default.
Which branch runs is decided per launch, not per tile, so any token count that is not a multiple of
`BN` sends the whole GEMM down the predicated path with a policy nobody chose.

Scope is that policy. No public type, workspace or dispatch threshold changes.

## Implementation

`kPredicatedCache` becomes a schedule parameter **defaulting to `Cache::ca`, so adding it changes no
existing instantiation**, and exactly one schedule opts in: `MmaR64C128`, the wide route of the linear
family. The opt-in goes through a member alias template so the trailing parameters, and their
defaults, cannot be respelled at the point of use. Two files, +32/−6.

**Six other call sites instantiate the same `<64, 128, 64, 16, 2, 2>` tile** — `linear_add`,
`gdn_input_proj`, `attn_input_proj` twice, `dflash2_attn_input` and `linear_pair`. They do not opt in,
so they compile to exactly what master has. Whether they would also gain is unmeasured and is a
separate change: the sign is not a property of the schedule. Forced onto `cg` the two BM=16 schedules
cost +16.8 % and +21.0 % of the operation, and it is not a property of `BM` either —
`MmaR64x16C48K128A1`, a BM=64 tile, gains 3.3 % on `[34816, 5120]` and loses 1.1 % on `[248320, 5120]`.

## Verification

### Numerical

A cache policy selects where a load is cached, not what it returns, so the output must be
bit-identical — witnessed rather than argued. `NINFER_OP_REPORT_STATS=1` at `%.17g` on both arms
across the six Op tests that reach this kernel:

```
ctest --test-dir build -j 4                                  # 118/118
NINFER_OP_REPORT_STATS=1 ./ninfer_linear_q8_a16_test          # 1099 records
NINFER_OP_REPORT_STATS=1 ./ninfer_linear_add_q8_a16_test      #  239
NINFER_OP_REPORT_STATS=1 ./ninfer_linear_pair_q8_a16_test     #  288
NINFER_OP_REPORT_STATS=1 ./ninfer_linear_swiglu_q8_a16_test   #  192
NINFER_OP_REPORT_STATS=1 ./ninfer_attn_input_proj_test        # 2824
NINFER_OP_REPORT_STATS=1 ./ninfer_gdn_input_proj_test         #  126
```

**4768 numeric records, byte-identical.**

"Changes no instantiation but the one that opts in" is checked in the object rather than asserted.
`cuobjdump -sass` on both builds: **22 function bodies and 334 `LDGSTS` on each side, and exactly one
body differs** — the predicated BM=64 BN=128 one, whose 15 `LDGSTS` go from 0 to 15 `BYPASS`, so the
file's total moves 167 → 182. Bodies are compared by content with addresses and encodings stripped,
because a changed template parameter changes every mangled name and `cuobjdump` re-aligns its comment
column to the longest one.

### Operator

`ninfer_linear_bench --qtype Q8 --execution graph --warmup 20 --repeat 300 --flush-mib 256`, three
passes with the arm order rotated, paired inside a pass, third arm the master binary again as an
identical-baseline null.

| shape and range | cells | median | worst | null p95 |
|---|--:|--:|--:|--:|
| feature 5120×25600, 129..255 | 10 | **−11.90 %** | −11.54 | 0.409 |
| generic 9216×2048, 4096..4223 | 122 | **−3.65 %** | −0.10 | 0.977 |
| mid 34816×5120, 129..319 | 11 | **−3.64 %** | −2.61 | 0.517 |
| lm head 248320×5120, 129..319 | 11 | **−3.90 %** | −3.12 | 0.847 |
| sparse 2048×4096, 896..1096 | 11 | −0.64 % | 0.00 | 0.602 |
| feature 50..56, a different kernel | 7 | **0.00 %** | 0.00 | 0.108 |
| feature 57..64, BM=16 | 8 | **0.00 %** | +0.37 | 0.527 |
| feature 65..127, BM=32 | 7 | **0.00 %** | 0.00 | 0.087 |

The last three rows are the routes this deliberately leaves alone, and they read zero within their
nulls. The number to carry is **−3.6 to −3.9 %** on the three large shapes whose grids run many
waves; the −11.9 % on the feature shape at 129..255 is a partial-wave effect, since that grid is 80
to 160 CTAs on a 170-SM part and does not fill the machine once.

The generic range covers all 128 `BN=128` alignment phases once each.

**The control this sweep contains**, re-taken on its own with eight passes because the wide sweep
left it only two. `full` needs `(T % BN) == 0`, so T=4096 is the one width the change cannot act on:

| T | 4092 | 4093 | 4094 | 4095 | **4096** | 4097 | 4098 | 4099 | 4100 |
|---|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| Δ % | −2.76 | −2.87 | −2.94 | −2.89 | **−0.12** | −2.91 | −2.79 | −3.00 | −2.92 |

−0.12 % at the untouchable width, inside its own 0.77 % null, against −2.8 to −3.0 % at all eight
neighbours. No passes discarded there.

Absolute performance of the resulting implementation:

| shape | T | µs | logical GB/s | TFLOP/s |
|---|--:|--:|--:|--:|
| generic 9216×2048 | 4096 | 777.120 | 144.5 | 199.0 |
| mid 34816×5120 | 319 | 761.056 | 282.3 | 149.4 |
| lm head 248320×5120 | 319 | 5338.050 | 283.4 | 152.0 |
| feature 5120×25600 | 255 | 439.680 | 352.4 | 152.0 |

## Limits

* **The sparse cell is not resolved.** It reads −0.64 % here and read exactly 0.00 % on an earlier
  base; both sit inside their own nulls. It is reported, not claimed.
* **193 of 579 passes were discarded** on the null gate across the wide sweep — a third. The card is
  noisy; the narrow control above discarded none of 72.
* The arms are separate binaries built from one tree, with a second run of the master binary as the
  identical-baseline null, rather than all routes in one process: the policy is a compile-time
  template parameter, so one binary would need a runtime switch that does not exist.
* Five of the dispatch cells reaching `MmaR64C128` are measured, not all of them.
* Every other predicated instantiation keeps `ca`, including ones separately measured to gain from
  `cg`. Moving them is a separate change with its own per-shape measurement.
* No end-to-end claim; this is an operator-level change.
* No Tensor Core utilisation is quoted beyond the benchmark's own TFLOP/s column.


MichaelDementii:
Supersedes #243, which was written against the W8 names before the Q8 rename. Scope is in #256.

@codex review

Neroued (approved):

PR #266, opened 2026-09-16
perf(ops): tune q5 a16 linear and residual projection routes

MOVIBALE:
# perf(ops): tune q5 a16 linear and residual projection routes

Maintainer-invited batch tuning of the Q5_G64_FP16 A16 row-split family, following #259–#262. One
subject: which kernel the Q5 projections use at a given token count. No weight-format, math, public
input-domain or numeric-standard change.

Sign convention used everywhere below: `latency_reduction_pct = (baseline_us − final_us) /
baseline_us × 100`, so **positive means the new route is faster**; `speedup_x = baseline_us /
final_us` is reported separately. All times are raw medians in µs, not re-rounded by hand.

## Scope and design

Seven public Ops were evaluated on one baseline (`5b4303c0`, master with #262 merged) and one branch:
five pure-Linear text shapes (1024×5120, 6144×5120, 7168×5120, 5120×6144, 5120×17408) and the two
resident LinearAdd shapes (5120×6144 for `attention/output` + `gdn/output`, 5120×17408 for
`mlp/down`). Two Vision shapes (1152×1152, 1152×4304) share kernels with the pure routes and were
kept as a regression gate.

Three changes, all inside the Q5 row-split family:

1. **LinearAdd at one token**: `{1,1} GemvResidual` merged into the exact-column split2 route. The
   baseline GEMV reached the same number only when it could stage the activation beside its weight
   buffers: for a 16-row, 2-stage block that is 34,816 B of x (17,408 × 2 B) plus 21,504 B of
   weights (16 × 2 × 32 uint4 nibble, 8 uint4 high, 2 uint4 scale), i.e. 56,320 B together against
   the 48 KiB static limit — so at K=17408 the kernel was instantiated with `kStageX=false` and every
   warp re-issued its global loads for the whole activation vector (N × K × 2 B ≈ 178 MB of logical
   loads per call). The two warps per output row of the exact-column split2 kernel do not depend on
   that staging and measured faster; the DRAM/L1-side attribution of the old kernel's cost was **not**
   measured, so none is claimed. The now-dead Q5 residual GEMV and the unused `kResidual` parameter
   of the shared GEMV core were removed, which is why the two fused Q5 consumers of that core
   (attn_input_proj, gdn_input_proj) appear as mechanical argument-position updates.
2. **LinearAdd above 512 columns**: whole 512-column waves plus a narrow tail. Measured on this host
   for the production 128-wide schedule: 512 columns cost 456 µs at k=17408 (320 blocks, 96% of the
   BF16 tensor peak) and 513 columns 934 µs, and the step repeats at 1025. A remainder of at most 192
   columns (measured to stay under that wave for every T in the band) now goes to the narrow routes; a
   wider remainder keeps the single wide launch. Two launches write disjoint column ranges of one
   output, each reading and adding the residual once, so Op semantics, the A16 zero-workspace contract
   and the one-complete-Op-per-graph rule are unchanged.
3. **Pure Linear**: the narrow-column MMA organization with the Store epilogue (the briefed direction),
   plus an exact one-column split4 launch. Their tables previously sent every T from 17 (or 25) into a
   128-wide tile pass, which at 1024×5120 is 16 row-blocks of work for an Op whose useful work at T=17
   is a sixth of that. New schedules `MmaR64C16`, `MmaR64C32S3`, `MmaR32C128` and
   `Split4C1_k{5120,6144,17408}`.

The wave size that sets the new upper bounds is a property of the measured instance, not a general
theorem: with the launch bounds in play this kernel gets two resident blocks per SM on this 170-SM
part, so one 128-wide launch is ~340 blocks and each schedule has a fixed row-block count per column
tile (16 for r64, 32 for r32). `r32c128` is therefore kept only while its grid still fits one wave —
1280 columns at 1024 rows, 256 at 5120 rows — and both boundaries were confirmed by measurement, not
derived from the block count alone.

## Seven-group completion

All seven groups carry a baseline, a measured conclusion and a final state. Classification uses a 2%
band: on this host two back-to-back runs of one build agreed within 0.2% and within 0.7% of the other
session, so a smaller move is reported as unresolved rather than as a gain.

<details><summary>Per-group final state (baseline → final, µs; positive = faster)</summary>

| Group | T=1 | T=4 | T=8 | T=512 | T=1024 | Points ≥2% faster / unresolved / ≥2% slower |
|---|---|---|---|---|---|---|
| Linear 1024×5120 | 13.568 → 7.392 (+45.5%) | 15.616 → 15.968 (−2.3%) | 29.952 → 17.664 (+41.0%) | 122.112 → 71.424 (+41.5%) | 122.592 → 85.248 (+30.5%) | 279 / 1 / 1 |
| Linear 6144×5120 | 19.712 → 18.176 (+7.8%) | 21.760 → 22.272 (−2.4%) | 57.088 → 46.848 (+17.9%) | 263.936 → 264.960 (−0.4%) | 399.840 → 399.104 (+0.2%) | 107 / 171 / 3 |
| Linear 7168×5120 | 21.760 → 21.632 (+0.6%) | 25.856 → 25.824 (+0.1%) | 52.000 → 50.944 (+2.0%) | 265.984 → 265.984 (0.0%) | 403.040 → 403.200 (−0.0%) | 102 / 178 / 1 |
| Linear 5120×6144 | 27.904 → 19.232 (+31.1%) | 24.000 → 23.776 (+0.9%) | 64.768 → 48.224 (+25.5%) | 161.536 → 161.536 (0.0%) | 322.816 → 323.296 (−0.1%) | 253 / 28 / 0 |
| Linear 5120×17408 | 62.688 → 44.800 (+28.5%) | 56.448 → 56.608 (−0.3%) | 160.832 → 111.872 (+30.4%) | 454.400 → 452.608 (+0.4%) | 904.992 → 906.496 (−0.2%) | 251 / 27 / 3 |
| LinearAdd 5120×6144 | 19.712 → 19.744 (−0.2%) | 26.336 → 25.728 (+2.3%) | 40.192 → 40.192 (0.0%) | 165.120 → 163.296 (+1.1%) | 325.408 → 325.408 (0.0%) | 5 / 276 / 0 |
| LinearAdd 5120×17408 | 60.896 → 46.848 (+23.1%) | 58.656 → 59.136 (−0.8%) | 85.248 → 85.440 (−0.2%) | 456.448 → 458.272 (−0.4%) | 911.104 → 923.392 (−1.3%) | 3 / 275 / 3 |

Largest single wins: LinearAdd 5120×17408 T=513 933.632 → 481.024 (+48.5%, 1.941×) and T=1025
1405.664 → 943.872 (+32.9%); LinearAdd 5120×6144 T=513 333.472 → 173.344 (+48.0%, 1.924×), T=1025
504.928 → 335.616 (+33.5%); Linear 1024×5120 T=18 128.288 → 21.024 (+83.6%, 6.102×); Linear
5120×17408 T=64 431.552 → 173.472 (+59.8%, 2.488×). Two further composite wins were confirmed by the
same-window A/B but are not in the 281-point curves: T=704 (tail=192) +11.0% on 5120×6144 and +16.7%
on 5120×17408.

Per-point data for all 281 token counts of each group: `final_vs_baseline_full.csv` (delivered with
the branch, see **Evidence**). The 35-row `final_vs_baseline.csv` is the key-point extract only.

The entries in the last column of the "≥2% slower" bucket are points on routes this PR does not
change; the same-window A/B in **Re-measurement** shows that none of them reproduces, so they are
cross-session drift. They are left in the CSV under the band rule rather than re-labelled by hand.
</details>

## Per-group results

<details><summary>Linear 1024×5120 — the largest single win</summary>

Routes: `1 → Split4C1`, `2..48 → SimtR8C4`, `49..160 → MmaR64C16`, `161..512 → MmaR64C32S3`,
`513..1280 → MmaR32C128`, `1281+ → MmaR64C128`.

The old table sent 17..∞ into a 128-wide tile: at 1024 rows that is 16 row-blocks per column tile, so
T=17..112 all cost a flat 128–132 µs. Measured now: T=17 21.3, T=18 21.024, T=20 21.408, T=32 34.0,
T=64 46.336, T=112 46.336, T=128 46.3, T=512 71.424, T=1024 85.248 µs. `r32c128` is bounded at 1280
because its grid leaves one wave at 11 column tiles. The switch itself was measured in a same-window
A/B against the baseline build: T=1280 122.112 → 87.808 µs (+28.1%), T=1281 140.544 → 140.544 (0.0%),
T=1282 140.512 → 140.544 (0.0%) — i.e. the 1280 point keeps the narrow-row win and 1281 costs exactly
what the baseline wide route cost, with no cliff at the boundary.
</details>

<details><summary>Linear 6144×5120 and 7168×5120</summary>

6144×5120: `1 → Split4C1`, `2..6 → ksplit` (unchanged), `7..11 → SimtR8C4`, `12..112 → MmaR64C32S3`,
`113+ → MmaR64C128`. Mid band 122–135 µs → 59–112 µs; T=112 134.656 → 112.384, T=113 136.448 →
135.648 (both sides of the switch, the latter unresolved).
7168×5120: same table. T=32 130.304 → 61.184 (+53.1%); T=113 135.936 → 135.904 (unresolved). The 113+
wide band is unchanged: measured `r32c128`/`c64` lose to `c128` from 128 up on both geometries
(6144×5120 T=192: 187.6 vs 145.2 µs), so no route was added there.
</details>

<details><summary>Linear 5120×6144 and 5120×17408</summary>

`1 → Split4C1_k`, `2..6 → ksplit` (unchanged), `7..15 → SimtR8C4`, `16..112 → MmaR64C32S3`,
`113..256 → MmaR32C128`, `257+ → MmaR64C128`. At 5120 rows the r32 grid leaves one wave at 2 column
tiles, i.e. 256 columns — measured at 288 columns `r32c128` is 235.3 µs against 171.7 µs for `c128`,
so the boundary sits exactly where the mechanism says. Mid band: 5120×6144 T=32 153.344 → 65.280
(+57.4%), T=256 146.880 → 110.336; 5120×17408 T=64 431.552 → 173.472 (+59.8%), T=256 404.736 →
304.896 (+24.7%). T=257 is unchanged on both (169.216 and 462.560 µs).
</details>

<details><summary>LinearAdd 5120×6144 and 5120×17408</summary>

Both shapes: the `{1,1}` GEMV interval merged into `{1,13}` / `{1,16}` split2, and `{193,∞}` C128 split
into `{193,512}` C128 + `{513,∞}` whole-wave-with-narrow-tail. Below 513 nothing else moved; measured
T=256/384/511/512 and T=768/1024 against a wide candidate set (`c16`, `c24`, `c32s3`, `c32s4`, `c64`,
`r32c128`) and none beats the current route beyond the same-session spread. T=1024 is a whole number of
waves with a zero tail, so it keeps the single wide launch: the same-window A/B measures it at +0.9% on
5120×17408 with one graph node, i.e. the composite is not entered there. The 64-column-and-wider narrow
columns re-read the weights per column tile — at T=1024, `c32s3` costs 2129 µs against 902 µs for
`c128` on 5120×17408 — which is why the wave composite sends only ≤192 columns to them, and why the
boundary is 192 rather than "everything above the last whole wave": at 193 columns the composite is
neutral (T=705 +0.1% / +0.3%) and keeps the single wide launch, while at 192 it still wins (+11.0% /
+16.7%).
</details>

## Numerical and boundary checks

Three different kinds of check were run at three different times. They are listed separately because
they are not interchangeable.

<details><summary>(a) What the development screening did — candidate vs. production, not an oracle</summary>

Candidate selection used a task-local probe (never registered in the repo's build) that ran each
candidate and the current production route on **identical buffers** and compared the two outputs
against each other: `maxdiff = 0` bit-exact for the pure-Linear candidates, and ~0.6% relative for the
pre-existing residual variants. That establishes "this candidate computes what the route it would
replace computes" — it is a consistency check against another implementation, **not** an independent
mathematical qualification. It was used only to avoid timing a candidate that computes something
else; the passing routes were then qualified against the FP64 oracle below.
</details>

<details><summary>(b) What the retained routes are qualified by — the existing public-Op suites</summary>

The routes this PR keeps are covered by the pre-existing public-Op harnesses, which compare against an
independent FP64 reference (sampled rows for most shapes, every row and column for 1024×5120) with
the pre-existing acceptance criterion — one criterion for the whole fused A16 Op, not selected per
route, per T or per shape, and not loosened here. LinearAdd additionally checks initial residual + W×X
with the reference inputs read before the update, verifies x/weight preservation, and the guarded
output buffers verify that a sliced launch writes no column outside its slice.
</details>

<details><summary>(c) What this review round added</summary>

* **Pure Linear.** `a16(1281)` and `a16(1282)` added to the 1024×5120 case, so the first values of the
  restored wide route after the 1280 bound are executed against the FP64 oracle (this was the gap
  flagged in review). Graph capture/replay was added on the routes this batch introduced or re-bounded,
  via the existing `Invocation` declaration: T=1 (split4 c1), T=17 (r64c16), T=512 (r64c32s3), T=1281
  (restored r64c128) for 1024×5120; T=1, T=12 (r64c32s3), T=1024 (r64c128) for 6144/7168×5120; T=1,
  T=16 (r64c32s3), T=128 (r32c128) for 5120×6144/17408 — 23 graph-replayed invocations in total
  (the pre-existing T=4/T=6 graph cases are kept). Under graph replay the Op is captured once, replayed
  twice, and both replays are compared against the FP64 oracle; the second replay negates the
  activation, which also catches a graph that ignores its input buffer.
* **LinearAdd.** The composite resolves its tail through the same route table, so its internal
  boundaries were covered by declaring the whole-wave offset (512) in the shape case: the harness then
  probes every route start again one wave later. The executed token set is now
  `1, 2, 3, 8, 13, 14, 15, 24, 32, 33, 34, 40, 48, 49, 50, 56, 64, 96, 128, 129, 192, 193, 194, 256,
  512, 513, 514, 525, 526, 527, 544, 545, 546, 560, 561, 562, 640, 704, 705, 706, 768, 1024, 1025,
  1026` for k=6144 and `... 528, 529, 530, 544, 545, 546, 560, 561, 562, ...` for k=17408 — i.e. the
  tail's split2→c16 switch (525/526/527 and 528/529/530), c16→c24 (544/545/546), c24→c32s4/s3
  (560/561/562), the composite→single-wide fallback at tail=192/193 (704/705/706), and a zero, one and
  two column tail after the second whole wave (1024/1025/1026). The list was confirmed by instrumenting
  the harness locally and dumping it, not inferred from the table.
* **Graph numerical checks for LinearAdd.** Both Q5 shape cases now declare `graph_tokens`
  `{513, 526/529, 545, 561, 705, 1025}`, so the composite is captured and replayed through the
  pre-existing graph branch (input update, residual restore, three replays including the X=0 identity
  replay, FP64 comparison each time). Previously only T=128 was graph-checked, so T≥513 was not.
* `q5_linear_add_workspace_capacity_bytes` still returns 0 for A16 across the new composite interval,
  and the bench reports `workspace_used=0` at every measured point (the A16 zero-extra-workspace
  contract is intact). Graph node count is 1 outside the composite and 2 inside it.

Suites run and passing after these additions: `ninfer_linear_q5_a16_test`,
`ninfer_linear_add_q5_a16_test`, `ninfer_attn_input_proj_test`, `ninfer_gdn_input_proj_test`,
`ninfer_gdn_input_proj_conv_snapshot_test`, `ninfer_gdn_input_proj_conv_record_test`,
`ninfer_gdn_replay_fold_test`. No sanitizer run was required: the production change adds no
allocation, no new buffer and no synchronization (the composite is one extra launch on the same
stream).
</details>

## Re-measurement of the switch points and the slower-than-baseline points

The first draft of this description reported a handful of points as slower than the baseline on routes
this PR does not change (Linear 1024×5120 T=4; 6144×5120 T=3/4/6; 7168×5120 T=3; 5120×17408 T=3/5/6;
LinearAdd 5120×17408 T=103/104/164 and T=1024). Those numbers came from two sessions hours apart. They
were re-measured in **one window with both builds alternating on the same GPU** — a separate worktree
at `5b4303c0` with its own build directory, the same benchmark sources on both sides, the same
protocol, three rounds each with the order reversed every round:

<details><summary>Same-window A/B/A, decision points (µs)</summary>

| Group | T | baseline | final | reduction | rounds (baseline / final) |
|---|---|---|---|---|---|
| L1 | 4 | 15.616 | 15.616 | 0.0% | 15.58/15.62/15.62 / 15.62/15.62/15.62 |
| L1 | 1280 | 122.112 | 87.808 | +28.1% | 122.11/122.11/122.62 / 87.30/87.81/89.34 |
| L1 | 1281 | 140.544 | 140.544 | 0.0% | 138.50/140.54/140.54 / 140.54/140.54/140.48 |
| L1 | 1282 | 140.512 | 140.544 | 0.0% | 138.50/140.54/140.51 / 138.50/140.54/140.54 |
| L2 | 3 / 4 / 6 | 19.712 / 21.760 / 28.416 | 19.712 / 21.760 / 27.904 | 0.0% / 0.0% / +1.8% | spreads ≤1.2% |
| L3 | 3 | 23.136 | 23.104 | +0.1% | 22.27/23.14/23.52 / 22.27/23.10/23.23 |
| L5 | 3 / 5 / 6 | 46.336 / 62.720 / 66.816 | 46.336 / 62.720 / 66.816 | 0.0% | spreads ≤1.6% |
| A1 | 16 | 63.264 | 64.512 | −2.0% | 62.94/63.26/64.19 / 64.51/63.23/64.74 |
| A1 | 513 | 333.568 | 173.472 | +48.0% | 333.28/333.57/333.57 / 173.41/173.47/173.54 |
| A1 | 704 (tail 192) | 347.296 | 309.024 | **+11.0%** | 347.30/347.90/346.91 / 308.70/309.02/309.31 |
| A1 | 705 (tail 193) | 347.584 | 347.392 | +0.1% | 346.91/347.90/347.58 / 347.62/347.39/347.26 |
| A1 | 1025 | 505.184 | 335.168 | +33.7% | 503.81/505.60/505.18 / 335.17/335.30/334.62 |
| A2 | 103 / 104 / 164 | 263.616 / 263.456 / 376.096 | 263.424 / 263.424 / 376.064 | +0.1% / 0.0% / 0.0% | spreads ≤0.3% |
| A2 | 513 | 933.504 | 474.464 | +49.2% | 933.50/933.34/933.60 / 474.14/474.46/475.90 |
| A2 | 704 (tail 192) | 965.888 | 804.608 | **+16.7%** | 964.35/966.37/965.89 / 801.31/804.61/804.61 |
| A2 | 705 (tail 193) | 964.352 | 961.056 | +0.3% | 964.35/964.35/966.40 / 961.06/960.26/961.79 |
| A2 | 1024 | 907.008 | 898.816 | +0.9% | 904.96/907.01/907.01 / 898.82/897.76/902.91 |
| A2 | 1025 | 1402.624 | 918.944 | +34.5% | 1400.90/1402.62/1402.62 / 916.96/918.94/923.65 |

</details>

What it establishes:

* **No regression survives.** Every point that looked slower in the cross-session comparison measures
  equal to the baseline in the same window (≤0.25%, inside each build's own round-to-round spread).
  They were session drift; `final_vs_baseline_full.csv` keeps them `unresolved` with both absolute
  times instead of counting them as gains, and the A1 T=16 case (whose round spreads overlap,
  62.94–64.19 vs 63.23–64.74) is likewise left undecided rather than explained away.
* **The 1280/1281 bound is confirmed from both sides**: the last r32c128 point is 28.1% faster than
  the baseline, the first restored wide point costs exactly the baseline time.
* **Two composite wins that the first draft did not claim**: A1 T=704 (tail=192) 347.296 → 309.024
  (+11.0%) and A2 T=704 965.888 → 804.608 (+16.7%). Their neighbours at T=705 (tail=193, the fallback)
  are unchanged (+0.1% / +0.3%) and the same run records the graph node count per point, which is 2 at
  513, 704 and 1025 and back to 1 at 705 and 1024 — the composite is entered exactly where the tail is
  in (0, 192] and nowhere else.

## Reproduction

Hardware and conditions, once: single RTX 5090 D (170 SMs, sm_120a), CUDA 13.1, Ubuntu 24.04 under
WSL2, driver as shipped; build type Release, `-DCMAKE_CUDA_ARCHITECTURES=120a`; one GPU measurement
queue at a time, correctness tests never run concurrently with timing. Timing protocol for every
number: one complete public Op per CUDA graph, L2 flushed with a 256 MiB buffer outside the timed
interval, 5 warmup launches, 50 measured samples, median, cold cache. For LinearAdd the residual is
restored to the same finite non-trivial value before every warmup launch and every sample, and the
restore and the flush are outside the reported time.

```bash
export PATH=/usr/local/cuda-13.1/bin:$PATH
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_CUDA_ARCHITECTURES=120a \
  -DNINFER_BUILD_APPS=OFF -DNINFER_BUILD_BENCHMARKS=ON
cmake --build build --parallel 6 --target \
  ninfer_linear_bench ninfer_q5_linear_add_bench \
  ninfer_linear_q5_a16_test ninfer_linear_add_q5_a16_test

# correctness: the two suites that cover the routes changed here
./build/tests/ninfer_linear_q5_a16_test
./build/tests/ninfer_linear_add_q5_a16_test

# pure Linear, one shape (repeat per shape with its own --n/--k)
./build/bench/ninfer_linear_bench --qtype q5 --policy a16 --n 1024 --k 5120 \
  --sweep 1:128:1 --execution graph --warmup 5 --repeat 50 --flush-mib 256 --csv-out lin.csv

# LinearAdd, one shape
./build/bench/ninfer_q5_linear_add_bench --k 17408 \
  --t-sweep 1,2,4,8,512,513,704,705,1024,1025 \
  --execution graph --warmup 5 --repeat 50 --flush-mib 256 --csv-out add.csv
```

`ninfer_q5_linear_add_bench` is the bench this PR teaches the graph/CSV mode: it times the public
`ops::linear_add` Op, reports the public shape, T, real graph node count and the public workspace
query, and no longer depends on private plan names. It is the only bench file changed here.

## Limitations

* **Which groups have a caller.** A1/A2 are on real paths (`attention/output` and `gdn/output` at
  `text.cpp:924`/`1059`, `mlp/down` at `ffn.cpp:87`). For the five pure-Linear text shapes this round
  found **no confirmed in-repo caller**: the fused `attn_input_proj` and `gdn_input_proj` have their
  own plans and never call `ops::linear`, and the MTP path uses Q8 for those geometries. Their
  improvement is real and reproducible at the public Op — the surface the Q5 linear benchmark suite
  uses — but it is not claimed as a model-level gain.
* **Operator, not model.** All numbers are per-Op latency, cold cache, one Op per graph. No
  whole-model throughput or generation evaluation was run, so nothing here should be read as a
  token/s claim.
* **Host spread.** The device reports a low but nonzero utilization when idle (6–22% sampled with
  nothing running). That floor is the desktop: on the Windows side ~30 processes hold GPU contexts and
  ~8.6–9.3 GB stays resident. This was measured, not assumed — closing two of the obvious candidates
  (ToDesk, Wallpaper Engine) freed 0.7–0.8 GB of memory and changed the utilization by nothing. With
  both of them actually gone later (ToDesk stopped, Wallpaper Engine's renderer exited, ~1.2 GB
  released, idle narrowed to 12–15%) the whole A/B/A driver was re-run on the quiet host: the
  baseline replays to a median 0.02% of the earlier session, the typical round-to-round spread on
  Ops ≥100 µs tightens from 0.29% to 0.07%, and no reduction quoted here moves by more than 3.3
  points. The load was jitter, not bias. So small deltas still need a control: the points that first
  looked slower were re-measured in a same-window A/B and none reproduced, and the surviving
  differences are reported as unresolved with their absolute times rather than counted as gains. The
  stability claim itself rests on the A/A and A/B runs (two back-to-back runs of one build agreed
  within 0.2%), not on this reading.
* **Not taken in this PR**: the same wave composite for the five pure-Linear shapes (they show the
  same structure at 385/513 columns and at 1408 for `r32`), because their routes have no confirmed
  caller and the budget went to verifying the three changes that do; and a K-split wide tile with a
  reduction for very long sequences, which would need a workspace and would change the A16
  zero-workspace contract — a design question for the maintainer rather than a tuning change.

## Evidence

The branch keeps only production code, the two test files and the one bench change. The measurement
data is delivered alongside it, not committed (per the earlier review comment about generated figures
and raw experiment output):

* `final_vs_baseline_full.csv` — 281 token counts × 7 groups, baseline and final medians, the
  reduction and speedup per point and the classification used above.
* `final_vs_baseline.csv` — the 35-row key-point extract quoted in this description.
* `data/` and `data_final/` — the raw bench CSVs the two above are built from. Each row carries the
  Op label, qtype, policy, N, K, T, timing mode (`execution=graph`), graph node count, warmup, repeat
  and flush size, so a reader can tell which protocol and which session a number came from; the
  session and clock conditions of each batch are in `baseline_r5.log` and `final_r5.log`.
* `baseline_r5.sh`, `final_r5.sh`, `compare.py`, `make_figures.py` — the exact commands that produced
  the curves, plus `remeasure_ab.log` for the switch-point re-measurement.
* `q5_r5_final_absolute.svg/.png` and `q5_r5_vs_baseline.svg/.png` — the final absolute curves and the
  baseline comparison.

These files are attached to this PR's conversation by the author (GitHub does not accept a local path
as an image link, and no upload URL is invented here). The equivalent raw data is reproducible with
the commands in **Reproduction**.


MOVIBALE:
Review round applied in `69e77c30` (tests only, no production change). What changed since the review:

* **The 1281 boundary is now executed.** `a16(1281)` and `a16(1282)` were added to the 1024×5120 pure-Linear case and 1281 is graph-replayed, so the first token counts of the restored wide route are compared against the FP64 oracle in both eager and graph mode. (This was the inline finding on `tests/ops/linear/test_q5_a16.cpp`.)
* **The composite's internal boundaries are covered.** The LinearAdd harness now declares the whole-wave offset (512) and probes every route start again one wave later, which is where the composite's tail resolves against the same table: `525/526/527` (k=6144) and `528/529/530` (k=17408) for the tail's split2→c16 switch, `544/545/546` (c16→c24), `560/561/562` (c24→c32s4/s3), `704/705/706` for the composite→single-wide fallback at tail=192/193, and `1024/1025/1026` for a zero, one and two column tail after the second whole wave. The executed token list was confirmed by instrumenting the harness and dumping it.
* **Graph replay now covers the routes this PR introduces.** 23 graph-replayed invocations across the five pure-Linear shapes (T=1 split4 c1, the narrow-column tiles, r32c128, the restored wide route) plus `graph_tokens {513, 526/529, 545, 561, 705, 1025}` on both LinearAdd cases, each captured once, replayed twice and compared against the FP64 oracle both times, the second with a negated activation.
* **The numeric evidence is described honestly now.** The development screens were candidate-versus-production comparisons (`maxdiff`, delivered in `data_linscreen/` / `data_cand/`) and are no longer described as oracle qualification; the retained routes are qualified by the existing public-Op suites against the FP64 reference with unloosened criteria; the additions above are listed as this round's work.
* **Report facts corrected**: the GEMV shared-memory budget (21,504 B of weight buffers + 34,816 B of activations = 56,320 B against the 48 KiB static limit — the old text read as if 56.3 KB were the buffers alone), the "≈178 MB of DRAM traffic" claim (that was arithmetic on logical loads, no counter was read), the wave-size statement (scoped to this 170-SM instance and its launch bounds rather than stated as a general theorem), the percent sign convention (positive = faster, one convention, regenerated from unrounded medians), and the per-point data claim (`final_vs_baseline_full.csv` is the 281-point table; the file previously named was a 35-row extract).
* **A same-window A/B/A** against a freshly built `5b4303c0` (same bench sources, three rounds, order reversed each round) resolves the points that had looked slower: none reproduces, so they were drift between sessions rather than regressions. The run also confirms the 1280/1281 bound from both sides (T=1280 +28.1%, T=1281/1282 0.0%) and measures the composite on both sides of its own boundary — **T=704 (tail=192) is +11.0% on 5120×6144 and +16.7% on 5120×17408**, while T=705 (tail=193, fallback) is unchanged, with the graph node count confirming which path ran.

Suites: `ninfer_linear_q5_a16_test`, `ninfer_linear_add_q5_a16_test`, `ninfer_linear_add_q8_a16_test` and `ninfer_linear_add_bf16_a16_test` all pass (the last two because the shared `ShapeCase` gained a trailing field — it is appended, not inserted, since the q8 case initializes that aggregate positionally).

Nothing here changes the production implementation; if the direction is acceptable the remaining question is only whether you want the same wave composite applied to the five pure-Linear shapes, which currently have no confirmed in-repo caller.

MOVIBALE:
Correction to the "Host spread" note in this description (now fixed in the body).

The text said this machine carries "a continuous background GPU load (~9.3 GB resident, 20–30% utilization)" and named desktop applications as the cause. That was wrong, and the user caught it: nobody was using the machine during the batch. Re-checked with the device idle — `nvidia-smi --query-compute-apps` returns **no compute process at all**, and six consecutive samples still read `6–22%` utilization, `~73 W`, `~9.3 GB` resident:

```
22 %, 74.73 W, 2077 MHz, 9320 MiB
12 %, 73.94 W, 2197 MHz, 9365 MiB
19 %, 73.70 W, 2002 MHz, 9354 MiB
13 %, 73.23 W, 2002 MHz, 9324 MiB
13 %, 74.21 W, 2070 MHz, 9380 MiB
 6 %, 62.97 W, 1695 MHz, 9348 MiB
```

So the readings are real but the interpretation was not: this device reports utilization from the Windows desktop's own composition (WDDM accounting), and the 9.3 GB is WDDM-resident desktop memory belonging to no process. I never enumerated processes before naming applications; that list was a guess and is removed from `RUN_RESULT.md`, `PR_BODY.md` and `PROGRESS.md`, with the raw correction written up in `REVIEW_FIX_RESULT.md` §8 of the delivered package.

Nothing measured changes: the stability argument was never based on that reading — it rests on the A/A and A/B runs (two back-to-back runs of one build agreeing within 0.2%) — and no reported number or threshold is affected. It was a host description I should have verified before writing, the same class of mistake as the two report facts corrected earlier in this round.

MOVIBALE:
Follow-up to my previous comment: that correction was itself partly wrong, and the user pushed back on it. Closing the two apps it names gave a clean measurement of what they actually contribute, so here is the corrected version.

**What I got wrong the first time.** I wrote "no compute process on the device" based on `nvidia-smi --query-compute-apps` — but that query was run **inside WSL**, which only enumerates CUDA applications in that VM. Run on the Windows side, the same query lists ~30 processes holding GPU contexts:

```
explorer.exe (x2), ShellHost.exe, ShellExperienceHost.exe, StartMenuExperienceHost.exe,
SearchHost.exe, TextInputHost.exe, CrossDeviceResume.exe, PhoneExperienceHost.exe,
PowerToys.exe + ColorPickerUI + FancyZones + PowerLauncher + Peek.UI, NVIDIA Overlay.exe (x2),
MyDockFinder\Dock_64.exe, steamwebhelper.exe, msedgewebview2.exe (x2), OneDrive.exe, chrome.exe,
ChatGPT.exe, codex-computer-use-swift.exe, WeChatAppEx.exe, Weixin.exe, QQ.exe, wallpaperui.exe
```

**What closing the two apps actually did** (measured, since the user asked for it):
* ToDesk — service stopped (one UAC elevation) and its processes gone: resident memory dropped from 9320–9380 MiB to 8558–8656 MiB, i.e. it was holding **0.7–0.8 GB**.
* Wallpaper Engine — **does not stay closed** on this install. The full tree returns within about 60 s of being killed, respawned as `wallpaper64.exe -silent` (it registers for Windows' application restart and has a Run entry at `HKU\...\CurrentVersion\Run`). It has to be exited from its own tray menu, or Steam closed.
* GPU utilization while idle: **6–22% before, 12–17% after** — i.e. neither app was the source of it.

So the honest wording, now in `RUN_RESULT.md` §6, `PR_BODY.md` and `REVIEW_FIX_RESULT.md` §8: an idle Windows desktop with this software installed keeps ~30 GPU clients and ~8.6–9.3 GB resident, and reports a low but nonzero `utilization.gpu` that does **not** come from Wallpaper Engine or ToDesk and cannot be attributed by that metric alone. It is a desktop floor, not a competing compute workload.

Nothing measured changes: the noise floor was established by the A/A and A/B runs themselves (two back-to-back runs of one build agreed within 0.2%), not by this reading, and no reported number or threshold depends on how the host is described. This is the third host/attribution sentence corrected this round, all three by measurement rather than by rewording.

Neroued (approved):

PR #271, opened 2026-09-17
perf(ops): route the q4/q5 a16 input projections by column band

MOVIBALE:
# Q4/Q5 A16 input projections: per-side routes for 1..12 columns

PR #271, revised after the directed review of the R6B result. Stages kept apart everywhere:

| label | state | meaning |
|---|---|---|
| **U** | upstream `f76e19c0` (#266 merged) | before this PR |
| **P** | `db88fa63` (R6) | stage one of this PR |
| **R0** | `3fa4a644` (R6B) | the T=7/8 row-block / K-split work |
| **R** | `99f63883` (this revision, head) | R0 plus the directed follow-up and the warp fence |

`U -> R` is the whole PR, `P -> R` the deeper-research gain, `R0 -> R` what the follow-up added.
Sign convention, used everywhere below: `latency_reduction_pct = 100*(old_us - new_us)/old_us`, so a
**positive** value means R is faster and a **negative** value means R is slower by that percentage;
`speedup_x = old_us/new_us` is reported as its own column. A duration that went up is written as a
negative reduction, never as a speedup.

The branch history was squashed into the single commit that is now the PR head, so the stage labels
above are the **pre-squash** commits retained as measurement provenance (`db88fa63` = P, `3fa4a644` =
R0, `8b6db778` = the T=9..12 band extension). The squashed commit's tree is byte-identical to the
pre-squash head, so every number below describes the same source content.

## What changed, and what stayed

R0 fixed the two parents' small-column kernels but only routed them at **T=7 and T=8**, on the
argument that from 9 columns on the two parents are already fused into one grouped kernel so a split
could not pay. That argument is wrong for attention - its resolver keeps the per-side form until
**13** (`cols <= 12 -> ParentSplit`, `cols <= 32 -> GroupedR32C32S2`) - and it was never checked as a
*complete call* for GDN. Measured as complete public ops, the split form is faster than the grouped
kernel at 9..12 for **both** ops and slower at 13:

* attention T=9..12: R0 100.1-105.7 us -> **73.0-77.6 us** (split with the K-split Q4 parent), grouped
  107.8-108.3 us; at T=13 grouped wins (107.8 against 129.8).
* GDN T=9..12: R0 grouped 120.1 us -> **101.6-105.7 us**; at T=13 grouped wins again (120.1 against
  126.2).

So R moves the per-side band from 1..8 to **1..12** in both ops (GDN's route catalog entry, and the two
launchers), with the Q5 side on its narrow 4-column SIMT tile in 9..12 - the shared-activation
row-block shape loses there (109.8-126.7 us) and an 8-column tile loses badly (163.1-165.1 us). The
4-column tile is the fastest of the three that were measured; the measurement does not single out one
cause for the difference. The strided/seam store epilogue now takes the problem's **actual** column
count rather than the tile capacity, so a future wider-capacity caller cannot write a masked tail
column.

**The 13..32 grouped band is not "unimproved by this PR".** Stage one (P) already improved it
substantially against U: GDN T=13 202.0 -> 123.9 us (**-38.7%**), attention T=13 114.2 -> 100.0 us
(**-12.4%**); T=16 and the wider extents show the same picture. What this PR does *not* do is extend
that work: 13+ keeps the grouped kernel exactly as P/R0 had it, and `R0 -> R` there is within
0.2% in either direction (see the resolution note below). Further grouped-kernel work is a future
direction, not a claim that the band was left alone.

## Final results on the public fused projections

Both ops, one session, stages alternating, cold graph, one complete public Op per timed interval,
median of 50, two interleaved rounds; every R row was re-measured in the same window as U/P/R0 after
the source fixes, so no comparison mixes two sessions (`RUN_RESULT_R6B.md` section 2, generator
`tools/summarize_pr.py`).

| T | GDN U | GDN P | GDN R0 | GDN R | GDN U->R | GDN R0->R | attn U | attn P | attn R0 | attn R | attn U->R | attn R0->R |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| 1 | 44.288 | 44.544 | 44.288 | 44.304 | -0.04% | -0.04% | 36.096 | 36.352 | 36.096 | 36.112 | -0.04% | -0.04% |
| 4 | 48.384 | 48.384 | 48.384 | 48.384 | 0.00% | 0.00% | 50.432 | 50.432 | 50.432 | 50.416 | 0.03% | 0.03% |
| 7 | 116.992 | 118.016 | 83.200 | 83.456 | **28.67%** | **-0.31%** | 85.248 | 85.248 | 64.768 | 64.768 | **24.02%** | 0.00% |
| 8 | 107.776 | 107.776 | 87.296 | 87.296 | **19.00%** | 0.00% | 77.056 | 77.056 | 66.816 | 66.816 | **13.29%** | 0.00% |
| 9 | 191.744 | 123.392 | 123.152 | **103.680** | **45.93%** (1.8494x) | **15.81%** | 109.824 | 109.824 | 109.824 | **75.264** | **31.47%** (1.4592x) | **31.47%** |
| 12 | 199.696 | 123.664 | 123.840 | **108.032** | **45.90%** (1.8485x) | **12.76%** | 109.824 | 109.728 | 108.992 | **80.384** | **26.81%** (1.3662x) | **26.25%** |
| 13 | 201.984 | 123.920 | 123.408 | 123.376 | 38.92% | 0.03% | 114.176 | 99.984 | 99.408 | 99.600 | 12.77% | -0.19% |
| 16 | 152.000 | 124.016 | 123.392 | 123.776 | 18.57% | -0.31% | 113.920 | 100.528 | 99.520 | 99.488 | 12.67% | 0.03% |
| 512 | 514.416 | 515.568 | 515.328 | 514.576 | -0.03% | 0.15% | 407.056 | 407.808 | 406.800 | 406.816 | 0.06% | -0.00% |
| 1024 | 911.104 | 911.840 | 914.080 | 915.728 | -0.51% | -0.18% | 801.536 | 801.264 | 803.072 | 799.856 | 0.21% | 0.40% |

Reading the small rows honestly: at T=1, T=4, T=13, T=16, T=512 and T=1024 nothing about the route
changed, and the entries there are **not resolved by this protocol** - the same R rows measured in a
second window move by up to ~4% in either direction (compare `data/pre_fence_R_*.csv` with
`data/R_*_r{1,2}.csv`), and the within-window two-round spread reaches 5.7 us at T=1024. GDN T=7 is
the one changed row that is not a gain: R0 83.200 -> R 83.456 us is **+0.26 us, i.e. 0.31% slower**,
which is inside that same uncertainty and is reported as unresolved rather than as either a cost or a
benefit.

## The complete Snapshot and Record pipelines inherit it

Same form, same (B, W), P against R, cold graph, one complete Op per timed interval:

| form | B | W | P us | R us | reduction | speedup |
|---|---|---|---|---|---|---|
| snapshot | 1 | 4 | 50.432 | 52.480 | -4.06% | 0.9610 |
| snapshot | 1 | 7 | 122.144 | 87.808 | 28.11% | 1.3910 |
| snapshot | 1 | 8 | 112.384 | 91.392 | 18.68% | 1.2297 |
| snapshot | 1 | 9 | 127.968 | 109.824 | 14.18% | 1.1652 |
| record | 1 | 7 | 122.112 | 87.296 | 28.51% | 1.3988 |
| record | 1 | 8 | 111.872 | 91.392 | 18.31% | 1.2241 |
| record | 1 | 9 | 128.000 | 108.288 | 15.40% | 1.1820 |
| snapshot | 2 | 4 | 109.824 | 89.344 | 18.65% | 1.2292 |
| snapshot | 2 | 8 | 126.240 | 127.744 | -1.19% | 0.9882 |
| record | 3 | 16 (valid 16\|9\|4) | 140.544 | 140.544 | 0.00% | 1.0000 |

That is also the direct evidence that R's projection is reached inside the complete Snapshot/Record
ops. `workspace_capacity` equals `workspace_peak` in every measured row. W=3/4/5/6 keep the same route
in P and R: repeating the B=1 pair in three separate windows moves those rows by **+-4%** of the P value
(`data/cd_b1_window_spread.txt`), so the table's W=4 -4.06% is that spread and is **not** read as a
regression; the W=7/8/9 rows are 5-7x that spread. The W=9 node column also differs structurally (P's
projection is one fused kernel, 4 nodes; R launches the two parents, 5 nodes) and R is still 14-15%
faster there.

Two labels in an earlier revision of this PR body were wrong and are corrected here: the numbers
"50.43 / 90.85", "87.49 / 126.14" and "91.39 / 126.21" are **snapshot B=1 against snapshot B=2** (the
same form at two batch sizes), not Snapshot against Record; and the statement that batch "does not move
the timed conv body at these widths" is false - W=4 is 50.4 us at B=1 and 90.8 us at B=2, because 4 and
8 aggregate columns take different routes.

## The racecheck warnings: what they are, and the fix

An earlier revision of this body said the racecheck warnings came from pre-existing code, that no
kernel this change adds appeared in any hazard, and that the report was therefore inconclusive. **That
was wrong, and it is replaced by the following.** (`compute-sanitizer` 2025.4.1.0 / build 37093031,
CUDA 13.1, driver 591.86.)

* **What the numbers are.** `4 hazards displayed (0 errors, 4 warnings)` is three
  `Race reported between ...` groups plus the tool's own `Maximum number of hazards reached` warning.
  It is three reported write-site groups covering twelve read instances, not four code locations, and
  the log is truncated by the hazard cap rather than exhaustive.
* **What the three groups are.** The `cp.async` copies of the nibble, high and scale planes
  (`cuda_pipeline_helpers.h:203` twice and `:190` once) against the previous slab's dequant reads
  (`q5_rowsplit_gemm_simt.cuh:76/77/79`). That header is a **shared** file - both small-T Q5 kernels
  call `dequant_chunk` and `q5_simt_issue_slab` from it - so the line numbers alone do **not** place the
  hazard in pre-existing code, which is exactly the inference the earlier text made.
* **Which kernel it is.** Re-running the same public op with the filter
  `--kernel-name kns=q5_rowsplit_rowblock` (no launch limit) reproduces the unfiltered report byte for
  byte, which places all three groups in
  `q5_rowsplit_rowblock_small_t_kernel<Q5RowSplitSimtSchedule, 8, 8, 2, true, 6144, Q5RowBlockStoreEpilogue>`
  - the row-block Q5 slab kernel **added by `3fa4a644` on this branch**, reached by both parents at
  T=7/8. Two controls show the filter discriminates: at T=7 the c4 SIMT instance is not launched and at
  T=9 the row-block instance is not launched, and both runs drop to millisecond medians with an empty
  report, so "analysed and clean" and "nothing ran" are distinguishable in the logs themselves (the
  benchmark prints its own median for the same call; the counter evidence is the byte-identical
  reproduction above, not the runtime).
* **The mechanism.** A warp-level **WAR**: the kernel issues slab `s+1` into the slot consumed at
  `s-1` as the first statement of the loop iteration, before any warp-level fence, and a 16-byte
  `cp.async` issued by one lane fills shared words that another lane reads in that previous iteration.
  The warp-per-row kernel of the same family carries `__syncwarp()` after its consume
  (`q5_rowsplit_gemm_simt.cuh:478`); the new kernel did not.
* **The fix, and the evidence it works.** One `__syncwarp()` after the consume loop in
  `src/ops/linear/q5/q5_rowsplit_rowblock_small_t.cuh` (one fence per slab, inside the uniform
  `full_slabs` loop, no early return before it), plus its comment. After rebuilding from those sources
  the same filtered runs - before at `data/racecheck_R_gdn_T7_rowblock.txt`,
  `..._T8_rowblock.txt`, `..._attn_T7_rowblock.txt`; after at `data/racecheck_afterfix_*.txt` -
  report **0 groups**, and the two "not launched" controls still behave as before, so the change is
  the empty report rather than a lost instrumentation. Per-site table, reachability matrix and the
  method (including why the first attempt at filtering reported a false "0 hazards") are in
  `RACECHECK_ATTRIBUTION.md`.
* **What this does *not* implicate.** The instance the T=9..12 band newly reaches
  (`q5_rowsplit_gemm_simt_kernel<Q5RowSplitSimtSchedule, 4, 8, 2, true, 6144>`) analyses clean in this
  build and in the upstream build at the same column counts, a catch-all Q5 filter at T=9 is clean, and
  the grouped kernel at T=13 is clean. So the extended band adds no exposure; the warnings were in code
  this PR itself adds.

## Verification

* Complete FP64 formula coverage of the changed projection mechanisms: the Q4/Q5 Snapshot pipeline is
  verified against the FP64 formula for **every output row and every state channel** at B=1 W=7/8/9 -
  the run log states it per case, e.g. `T=9 ... complete FP64 reference, compared 92160 output values
  + 276480 state values ... failures 0`. W=1..6, 32, 33, 64 keep the sampled reference, plus a batched
  mixed-valid-columns case (W=16, B=3, valid `16|9|4`). Note that the batched case is routed **once**
  from its aggregate column count (48), so it exercises mixed valid prefixes and exact-zero invalid
  tails, not three different per-row routes; the per-row coverage comes from the per-side suites.
* Record keeps its bit-level relationship with Snapshot, reads only the source state, and covers
  W=2..16 at B=1 plus ragged B=8 cases whose aggregate columns reach into the new band.
* Fused-op suites: GDN covers every route boundary of the new catalog (1..13, 15..17, 32, 33, 64, 65,
  127..129, 193) with graph replays at 7/8/9/12/13/33/65/129; attention covers the Q4/Q5 projection at
  **every** column count 1..128. The graph-replay check re-poisons and re-verifies each replay.
* The four affected suites were re-built from the fenced sources and re-run (`data/after_fix_tests.txt`,
  `data/test_snapshot_fullref.txt`, `data/test_record.txt`, `data/test_gdn_proj.txt`,
  `data/test_attn_proj.txt`), and memcheck / synccheck remain at 0 errors on both parents and on the
  complete Snapshot op.
* The Snapshot/Record benchmark now reports `graph_nodes` **per row**: eager rows report 0, and a
  captured row reports the whole graph including its two timing event nodes, so that column is not the
  number of production kernels (`data/both_mode_nodes.csv`).
* No oracle, tolerance or reference was changed, and the whole-repo suite was not re-run: only the four
  suites affected by these routes were, and they pass.

## Future directions, not implemented here

* A dedicated fused kernel for 7/8 columns: **not implemented**. No same-W candidate comparison was
  made for it, and the cross-width equality between W=3 and W=4 is not a substitute for one - the two
  widths happen to cost the same in the existing routes.
* Further work on the grouped kernel for 13..32 (stages, tile shape, jobs): not attempted in this
  round, and no measurement here supports a single-factor explanation of its ~120 us. Recorded as the
  remaining open limit, with the band's stage-one gains kept.
* A cache-policy or timer refactor in the repository: not attempted; the L2-eviction variants remain a
  measurement-conditions diagnostic (three protocols, not three kernels) and are **not** counted as an
  optimisation of this change.

## Evidence bundle and its delivery status

`evidence_r6b_followup.zip` (253 entries, ~697 KiB) contains `FINAL_REVIEW_RESULT.md`,
`RACECHECK_ATTRIBUTION.md`, `RUN_RESULT_R6B.md`, `RESEARCH_NOTES.md`, `ENVIRONMENT.md`,
`commands.txt`, `fix_round_diff.patch`, `r6b_diff_p_to_r.patch`, `r6b_diff_u_to_r.patch`, the raw CSVs
and logs under `data/` (including the original capped racecheck logs, the filtered runs before and
after the fence, and `data/bisect_*.txt`), the four figures under `data/figures/`, the generator
scripts under `tools/`, and the previous round's bundle.

**Delivery status, stated exactly:** the file was generated and handed to the author as a local file
(`C:\Users\MIN\Downloads\evidence_r6b_followup.zip`, same file also in the working tree at
`profiles/bench/input_proj_r6b/`). This environment has **no** way to attach a file to a GitHub PR, so
the bundle is **not attached to this PR and no upload URL exists** - it awaits upload by the author or
maintainer. External reviewers currently have the source, the review comments and the tables quoted in
them; they have **not** received the ZIP body and are not claimed to have re-run it or checked the logs
themselves.

## Merge status

Live GitHub state at the time of writing: `mergeable = MERGEABLE`, `mergeStateStatus = BLOCKED`,
`reviewDecision = REVIEW_REQUIRED`, base `f76e19c0` (unchanged), head `99f63883` (the squashed
single commit). `BLOCKED` here is the
review gate, **not** a source conflict - there is nothing to rebase and none was performed, no branch
protection was touched and no check was bypassed. No automatic merge.


MOVIBALE:
## Final review round: the race-check warnings are attributed, fixed, and re-verified

Follow-up on the review's one substantive question — *what do the four racecheck warnings have to do with this PR's production calls?* — plus the small fixes that were asked for. New commit: `b3610486` on `perf/q4q5-input-projections-r6`. No merge, no release.

### 1. What the four warnings are

`4 hazards displayed (0 errors, 4 warnings)` is **three hazard groups plus the tool's own "maximum number of hazards reached" warning** — not four sites. All three groups are in the row-block Q5 slab kernel **this PR adds** (`3fa4a644`):

* kernel instance (from the device frame of the detailed report):
  `q5_rowsplit_rowblock_small_t_kernel<Q5RowSplitSimtSchedule, 8, 8, 2, true, 6144, Q5RowBlockStoreEpilogue>`
* the two ends: the `cp.async` copies of the nibble/high/scale planes (`cuda_pipeline_helpers.h:203` twice, `:190` once) against the previous slab's dequant reads (`q5_rowsplit_gemm_simt.cuh:76/77/79`). That header is **shared** by both small-T Q5 kernels, so the capped log names a file and not a kernel — the kernel had to be isolated by filtering.
* type: warp-level **WAR**. The kernel issues the copies for slab `s+1` into the slot consumed at `s-1` as the first statement of the loop iteration, before any warp-level fence, and a 16-byte `cp.async` issued by one lane fills words another lane reads. The warp-per-row kernel of the same family carries `__syncwarp()` after its consume (`q5_rowsplit_gemm_simt.cuh:478`); this kernel did not.

Isolating it: `--kernel-name kns=q5_rowsplit_rowblock` reproduces the unfiltered report **byte for byte**; the c4 SIMT instance that the T=9..12 band newly reaches analyses **clean**, in this build and in the upstream build at the same column counts; the grouped kernel at T=13 is clean; and the two "instance not launched" controls (T=7 with the SIMT filter, T=9 with the row-block filter) drop to millisecond medians, so the "analysed and clean" reads above are not "nothing ran" (the benchmark prints its own median in the same log: 1.7-3.3e6 us when analysed).

### 2. What changed and what was re-verified

One line of production code: **`__syncwarp()` after the consume** in `src/ops/linear/q5/q5_rowsplit_rowblock_small_t.cuh` (one fence per slab, inside the uniform loop, no early return before it), plus its comment. Non-performance fixes in the same commit: the Snapshot/Record benchmark reports `graph_nodes` **per row** (eager rows 0 — verified in both-mode CSV); the GDN K-split store takes the problem's **actual column count** (`x.ne[1]`) instead of the tile capacity, the same number for every dispatched case today; and the stale explanations were corrected (catalog band, W=9 Q5 mechanism, the mixed-valid-columns case, the c8-vs-c4 note now reports the measured ranking without singling out a cause).

After the fence: the filtered re-run is **clean at T=7 and T=8 on both parents**, still instrumented; the four affected suites were rebuilt from these sources and pass (`ninfer_gdn_input_proj_test`, `..._conv_snapshot_test` with the complete FP64 reference at W=7/8/9, `..._conv_record_test`, `ninfer_attn_input_proj_test`); memcheck and synccheck stay at 0 errors. The fence costs **0.00-0.31%** at the extents it covers, which is below the protocol's spread.

### 3. Numbers

Because R's binary changed, the R column was **re-measured in the same window as U/P/R0** rather than patched, so no comparison mixes two sessions. Final, one window, cold graph, median of 50, one complete public Op per timed interval:

| row | U | P | R0 | R | U->R | R0->R |
|---|---|---|---|---|---|---|
| attention T=9 | 109.824 | 109.824 | 109.824 | **75.264** | **-31.47%** (1.4592x) | -31.47% |
| attention T=12 | 109.824 | 109.728 | 108.992 | **80.384** | **-26.81%** (1.3662x) | -26.25% |
| GDN T=9 | 191.744 | 123.392 | 123.152 | **103.680** | **-45.93%** (1.8494x) | -15.81% |
| GDN T=12 | 199.696 | 123.664 | 123.840 | **108.032** | **-45.90%** (1.8485x) | -12.76% |
| attention T=7 / T=8 | 85.248 / 77.056 | 85.248 / 77.056 | 64.768 / 66.816 | 64.768 / 66.816 | -24.02% / -13.29% | 0.00% / 0.00% |
| GDN T=7 / T=8 | 116.992 / 107.776 | 118.016 / 107.776 | 83.200 / 87.296 | 83.456 / 87.296 | -28.67% / -19.00% | -0.31% / 0.00% |

Complete Snapshot/Record ops at the same B/W, P vs R (B=1): snapshot W=7 **-28.11%**, W=8 **-18.68%**, W=9 **-14.18%**; record W=7 **-28.51%**, W=8 **-18.31%**, W=9 **-15.40%**. W=3/4/5/6 keep the same route in P and R and move within **+-4%** of the P value — that spread is measured by repeating the B=1 pair in three windows, so those rows are read as "no change", not as a regression. Sub-0.6% entries elsewhere are at or under the protocol's own two-round spread.

### 4. Evidence bundle

The bundle is a real file, not a directory name: `profiles/bench/input_proj_r6b/evidence_r6b_followup.zip` in the working tree, with a copy handed to the user at `C:\Users\MIN\Downloads\evidence_r6b_followup.zip`. This environment cannot attach files to a PR, so it is delivered as a file for the maintainer to upload; nothing is claimed to be attached here. It contains `FINAL_REVIEW_RESULT.md` (the closing answers), `RACECHECK_ATTRIBUTION.md` (per-site table, reachability matrix, the fence and its re-run), the small-fix diff (`fix_round_diff.patch`), the P->R and U->R patches, the raw CSVs and logs (including the capped original racecheck logs, the filtered runs before and after the fence, and the failed first filtered attempt), the four suites' logs, the figures, and a short environment note. Pre-fence R rows are kept as `data/pre_fence_R_*.csv`.

### 5. Status

**Suitable to merge as it stands; no unresolved blocker.** Two items stay out of scope and are stated as such rather than presented as done: the 13..32 grouped band is not improved by this PR, and the dedicated fused kernel for 7/8 columns is **not implemented** (no same-W candidate comparison was made for it; the cross-width W=3-vs-W=4 equality is not a substitute).

Neroued (approved):

PR #287, opened 2026-09-19
[2/5] feat(artifact): carry one NVFP4 divisor per source matrix in a stacked tensor

MichaelDementii:
Related Issue: #283

Second of five, and the only product-contract change in the set. Independent of the first; the third
needs both. No performance claim.

## What changes

A tensor object holds one NVFP4 weight divisor. That is right for a plane quantised as one matrix and
wrong for a plane assembled from several quantised apart — which is what every published NVFP4
checkpoint of an MoE model is: each expert matrix carries its own `weight_global_scale`, 30 720 of
them on the routed banks of `qwen3_6_35b_a3b`.

The object gains an optional `divisors`, the number of separately quantised source matrices stacked
into the plane; the layout and the row-to-word rule are on the Issue. Default 1.

## Contract

`divisors` absent means one word and today's layout exactly, so the stored form of every existing
artifact is unchanged. The reader is not: `validate_nvfp4_weight` gains two refusal conditions that
the six shipped dense NVFP4 wrappers run through, so a `Weight` built outside `native_weight` now has
to carry a divisor pointer. That is why two in-tree fixtures move with it —
`tests/ops/quantized_weight.h` and `bench/ops/quantized_weight.cuh`.

Four refusals in total: `divisors` outside NVFP4, a count that does not divide the rows, a share that
is not a whole number of scale tiles, and sources with unequal row counts. Two of them are the C++
conditions above; all four are enforced on the Python side. The count itself is decided where the
parent's shape is chosen, in the recipe plan, rather than being discovered at encode time.

The documents that own the rule move with it: the `nvfp4` row, the reconstruction equation, and the
producer and reader obligations of `docs/maintainer/tensor-formats.md`; the `payload_bytes` formula
in `docs/maintainer/storage-layouts.md`; the object-member table of
`docs/maintainer/artifact-container.md` §5.1.

`validate_nvfp4_weight` also gains a defaulted `stacked` flag that every route in this commit leaves
false, so a multi-divisor plane is refused everywhere until the route that can read one arrives.
Nothing in the engine reads more than one divisor yet; that is the next commit, and it is why this
one is separate.

## Verification

`pytest tests/artifact tests/convert`: 53 at this commit against 47 on master. The six new
ones are the stacked round trip, the collapse to a single divisor when the sources agree, the
128-row tile rule, the refusal outside NVFP4, and two on the activation divisor: that an `AllowA4`
input with no override carries the source's own calibrated word, and that a stacked parent takes the
smallest of its sources'. All six are synthetic; the end-to-end conversion of a vendor
checkpoint lands with the recipe, in the last commit.

`ctest` 120 of 120 — the C++ refusals and the multi-word read ride on it rather than on a test of
their own.

## Checks not run

* No C++ test constructs a plane with more than one divisor. The multi-word read, the per-word
  validation and the `stacked` branch are exercised only from the commit that adds a reader.
* No existing artifact of another architecture was loaded again to confirm the two new C++ refusal
  conditions are inert for it.
* Nothing measured: the divisors are read once per row inside an existing epilogue, and timing
  belongs to the commit that adds the reader.


MichaelDementii:
This one adds a `divisors` member to the NVFP4 tensor record so a stacked parent can keep one
per-source divisor instead of one for the whole plane, and it is the PR I am least attached to: if
you would rather have the rescale at conversion time, this withdraws and [5/5] changes with it.
Worth checking closely: that an artifact written before this change still binds (the member defaults
to 1 and the index collapses to word 0), and the activation-divisor rule in `tools/convert`, which
gates on the number of *sources* rather than the number of divisors — the other way round silently
breaks the two merged 27B recipes, and there are now two tests standing on exactly that.

@codex review

PR #289, opened 2026-09-19
[4/5] feat(engine): accept NVFP4 sparse-MoE expert banks

MichaelDementii:
Related Issue: #283

Fourth of five. Thirty lines, and the point at which the profile becomes reachable. It needs the
third. No performance claim.

## What changes

`prepare_sparse_moe_weights` enumerates the admissible bank formats and had no NVFP4 arm, so an
artifact whose experts carry the format could not be bound even once the operation could run it.

The profile takes all four expert matrices as NVFP4. They are inseparable: the gate/up epilogue emits
an encoded intermediate that `down` reads, so a mixed bank has no route.

It also requires `AllowA4` on every expert input. The prefill route quantises the hidden state to four
bits, which is exactly what that permission grants; every other NVFP4 consumer refuses an input
without it, and this one would have quantised the activation of an input whose policy says it must
not, and said nothing.

## Contract

A user who does not grant `AllowA4` now gets a refusal at bind time naming the reason, where before
the bank was rejected as an unsupported format. No existing profile's admissibility changes.

## Verification

`ctest` 120 of 120.

End to end: the published NVFP4 artifact loads and serves with this commit and is refused without it.
The serving numbers for the whole series are on the Issue.

## Checks not run

* **No unit test covers either the new format arm or the `AllowA4` refusal.** The only in-tree caller
  of `prepare_sparse_moe_weights` outside production is the real-artifact loading test, which needs
  an artifact this suite does not carry and therefore skips. The evidence is that the artifact binds
  with this commit and not without — weaker than a negative case.
* Nothing measured: this is a bind-time check.


MichaelDementii:
Small one: the engine's expert bank learns to admit NVFP4 planes. The question worth asking is
whether the admission predicate can accept a plane the prefill and decode routes cannot actually
consume — the format check, the divisor count and the scale-layout expectation have to agree, and
they are written in three places.

@codex review

PR #290, opened 2026-09-19
[5/5] feat(convert): build the published Qwen3.6-35B-A3B NVFP4 checkpoint into an artifact

MichaelDementii:
Related Issue: #283

Last of five. It needs the third; it is what produces the artifact the series exists for.

## What changes

A recipe, `qwen3_6_35b_a3b_nvfp4`, that builds the published NVFP4 releases of this model into an
artifact: routed and shared experts imported as they are stored, projections Q8, vocabulary Q8/Q6 —
the same assignment as the groupwise recipe everywhere except the experts.

The expert banks are grouped explicitly rather than by the automatic rule, and only in the text
component. Automatic grouping still separates parents whose divisors differ, and that is right: its
result is also read by dense routes whose kernels take one scalar, and only the MoE kernels can read a
stack. An explicit group is the statement "this one is a stack".

## Contract

A new recipe name in `docs/weight-conversion.md`; no existing recipe changes. Conversion needs no
external script and rewrites no block scale, so the artifact holds the vendor's own codes and scales
byte for byte.

Reproducing it needs three files the release omits and the loader requires: `preprocessor_config.json`
and `video_preprocessor_config.json`, taken from the BF16 release, and a `tokenizer_config.json`
carrying the `added_tokens` array of the release's own `tokenizer.json` — the loader cross-checks the
two, so an abbreviated one is refused.

## Verification

`pytest tests/artifact tests/convert`: 55 at this commit against 47 on master; the two new ones
cover the source mapping for a routed and a shared expert, and its refusal on a missing one.

`RedHatAI/Qwen3.6-35B-A3B-NVFP4` converts in one command; the artifact is 21 759 564 036 bytes against
the groupwise control's 22 244 450 816. Quality and the perplexity control are on the Issue.

## Checks not run

* No test runs the recipe itself — the new tests cover its source mapping, not the assignment, the
  `AllowA4` policy or the grouping rule.
* The recipe does not check at build time that a release's per-expert activation divisors agree.
  They do in the RedHatAI release, every one 131.0; a release where they differ would convert and
  then be refused at bind. Naming the tensor at conversion time would be better and is not done.
* Only the RedHatAI release was converted. The nvidia and unsloth releases carry per-expert divisors
  too; unsloth additionally stores its projections as `fp8_e4m3fn_row_bf16` at K=2048, a shape
  `src/ops/linear/fp8/fp8_shapes.h` has no family for.


MichaelDementii:
Last of the five: the conversion recipe for Qwen3.6-35B-A3B on NVFP4. It depends on [2/5] and
withdraws with it. Two things are worth checking against the published checkpoint rather than
against the recipe's own intent: that every stacked parent really does carry one divisor per source
expert (the whole point — collapsing them costs 63 % of the perplexity gap), and that the shared
expert, whose planes collapse to a single divisor in the real artifact, still takes the same path.

@codex review

